PR #100, opened 2016-01-26
you may want to wordsmith this a bit...

mrose17:
…or not, as the case may be!  @garvankeeley - enjoy!


mrose17:
actually, the `click "Choose"` part is, sadly, correct. even though i have a single developer identity with my appleID, it still asks me to pick 1 of the 1 choices. go figure...

PR #122, opened 2016-02-08
use new oip endpoint

bridiver:


garvankeeley:
Thanks, I'll try it out now to make sure I see replacements on time.com

Issue #245, opened 2016-05-30
Add the ability to re-order bookmarks


bbondy:
Requested from support.

diegodoval:
Also, adding 'remove'
![image](https://cloud.githubusercontent.com/assets/420063/17152440/a78bafd8-532c-11e6-87f9-ca9ce76bbaf6.png)

diegodoval:
Will close this as-is in the branch and then use the data in #301 to finish.

diegodoval:
Final spec described in #301

PR #306, opened 2016-07-27
(wip) Issue #245: re-order bookmarks

diegodoval:


brave/src/webview/PrivateBrowsing.swift
```diff
@@ -114,7 +114,7 @@ class PrivateBrowsing {
 
         getApp().tabManager.removeAllPrivateTabsAndNotify(false)
         delay(2) {
-            Answers.logCustomEventWithName("PrivateBrowsing exit failed", customAttributes: nil)
+//            Answers.logCustomEventWithName("PrivateBrowsing exit failed", customAttributes: nil)
```
garvankeeley: ``` #if !NO_FABRIC ``
Is needed to catch if PB mode exit failure rates suddenly shoot up after an iOS update.

Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -104,23 +108,47 @@ class SiteTableViewController: UIViewController, UITableViewDelegate, UITableVie
...
+        
         tableView.delegate = nil
+        
+    }
+    
+   
+    func switchTableEditingMode() {
+        //unwoned self is generally unnecessary here since the block is not going to create retention loops,
+        //but useful to include considering UIViews may get deallocated unexpectedly
+        dispatch_async(dispatch_get_main_queue()) { [unowned self] in
```
garvankeeley: Agreed on the weak, no comment needed here, maybe instead we should be commenting why strong binding is required, I lean towards
``` [weak self] in
self?.doStuff ``` 
Any reason not to?

File: brave/src/frontend/sidepanels/MainSidePanelViewController.swift [hunk 3]
```diff
@@ -53,6 +56,16 @@ class MainSidePanelViewController : SidePanelBaseViewController {
         settingsButton.addTarget(self, action: #selector(onClickSettingsButton), forControlEvents: .TouchUpInside)
         settingsButton.accessibilityLabel = NSLocalizedString("Settings", comment: "Accessibility label for the Settings button.")
 
+        
+//        editBookmarksButton.setTitle("Edit", forState: .Normal)
+        editBookmarksButton.accessibilityLabel = NSLocalizedString("Edit Bookmarks List", comment: "Accessibility label for the Edit Bookmarks List button")
+        let img = UIImage(named: "editBookmarks")
+        editBookmarksButton.setImage(img, forState: .Normal)
+        editBookmarksButton.addTarget(self, action: #selector(onEditBookmarksButton), forControlEvents: UIControlEvents.TouchUpInside)
+//        editButton.sizeToFit()
```
garvankeeley: Commented out code might be useful for debugging, but if not needed it can be removed

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -91,6 +105,13 @@ class MainSidePanelViewController : SidePanelBaseViewController {
             self.history.refresh()
         }
     }
+    
+    @objc func onEditBookmarksButton() {
+        NSNotificationCenter.defaultCenter().postNotificationName(BookmarksNotifications.SwitchEditMode, object: self)
```
garvankeeley: Can this be callback delegated, I want to keep notifications for the times we are desperate to connect disparate entities (I am just skimming the PR, so I don't recall how closely related these classes are).

Client/Frontend/Widgets/SiteTableViewController.swift
diegodoval: They both are equal as far as solving recurring references. The difference between weak and unowned is that weak creates an Optional while unowned implies the reference will never be nil, therefore causing a crash if nil. Since *in theory* you shouldn't have a nil reference the point of making it unowned is forcing a semantic view where the thing *should* crash if nil, allowing us to solve the problem. It's like an assert in this sense. Marking it as 'weak' means that a  problem like that would literally never be found since weak marks the variable as optional.

garvankeeley: Not an assert though, it is a crash. Why crash unless the app state would be dangerously broken, or non-recoverable? Particularly in UI code, often just moving to another screen or performing another action can repair the UI state. Optionals (or message-to-nil in obj-c) is fabulous for built in safety against edge-case UI breakages.

garvankeeley: BTW, I only skimmed this code without reading in detail, unowned self is likely fine here, just raising points for us to hammer out as we go forward in future. I'll have to read the code in more detail, I am not sure just yet why the dispatch to main is required here

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
diegodoval: There is in fact a reference to use, but I think this is a perfect example of when that reference shouldn't exist at all. It's a component that is parallel and neither "above" or "below" the other in the hierarchy. So to use a direct function calls means we have to propagate references to other things through various parts of the stack. (I'm just adding this here for clarity in case someone or us land here out of context).

Thing is, early on it wasn't clear _where_ to place the button, and I assumed it might have to be moved (which turned out to be true). The button right now is placed temporarily in the top toolbar and has to be moved into another toolbar at the bottom, using the notification seemed the simplest way of knowing that you can just take the button and put it somewhere else, not having to pass a bunch of references to that other toolbar, which could be created entirely self-contained. 

The nature of this function (and others) but focusing on this one, is that "someone requested we enter edit mode in the bookmarks list". It's irrelevant who requested that, and to the requester (in this case a button in a toolbar somewhere) it's irrelevant who is doing something about it. Using a notification, or really what it is (regardless of the name in Foundation), an asynchronous event, is a good way to avoid reference loops when they shouldn't even be a thing. And it means you can move UI elements around the hierarchy without having to pass through a lot of references that you need for this type of thing.

Should it be required that the button reacts by changing state, that can be handled by another message. 

From what I can see the hierarchy of visual classes has references passed up and down the stack, making it very difficult to do refactoring or modify its structure without breaking lots of things, so that's why I thought it was a good starting point for uncoupling. There's other advantages (for example, speed of development, since there's far fewer things you need to make a button -> action connection work) but this is a key point. 

All that said, if there's some specific concern (or you simply don't want to use async events) it'll be trivial to undo and use the reference. This is just my case for why it should be done this way. :)

Client/Frontend/Widgets/SiteTableViewController.swift
diegodoval: The issue as far as a design question is that regarding the object lifetimes, this should _never_ crash (I mean an assert in terms of visibility). So what we're expressing with unowned is that. The crash is a side effect of the components doing something we don't want. Using an optional means that we'd never see that problem. 

Here's my argument: 

Optionals are used when we know that in some cases the var will be nil. Swift's whole obsession with nullability stems fixing the badly defined behavior of NULL in Obj-C, or, really, in C. (eg, does NULL mean "something went wrong and this should exist" or does it mean "this is an empty pointer but I'm ok with that")

If we ask: "When do you _expect_ this variable to be nil?" and the answer is "never" then it should not be an Optional in Swift's worldview. Given that as far as we know this cannot happen, then we have no idea what the app state is if it does.

Using _weak_ would require us to wrap that in an _if let_ and then have an else clause logging or warning somehow about what is clearly a catastrophic error for the app's state. (Again, since we don't know how it can happen it's very likely that after this error other bad things would happen also leading to a crash). 

I prefer unowned in this case, but for sure I'd argue strongly against just using weak without testing it, it is something that we should know about if it happens (we should probably analyze the view tree in detail anyway at some point to have a clear picture of who owns who)

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
diegodoval: sizeToFit() is necessary I think, otherwise you need to set the frame of the button some other way. I still haven't quite gotten my head around the complex compositing system used here, but it's only been a few hours :)

Client/Frontend/Widgets/SiteTableViewController.swift
garvankeeley: My point is based around the argument about crash vs. not crash. I am completely on the no-crash side. 2 crashes and the user might delete the app.

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -91,6 +105,13 @@ class MainSidePanelViewController : SidePanelBaseViewController {
...
+        dispatch_async(dispatch_get_main_queue()) { [unowned self] in
```
garvankeeley: dispatch_async shouldn't be needed here, is all main thread UI code, no?

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
garvankeeley: OK lets go with this, NSNotifications are useful for this toolbar UI code (we inherited the dual-toolbar design complexity from Firefox iOS, so in some cases we lack a 'right' way to do things). I agree with your thinking on cleaning this up, NSNotifications can simplify decoupling, so we might do more of it in this case.

Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -104,23 +108,47 @@ class SiteTableViewController: UIViewController, UITableViewDelegate, UITableVie
...
+        }
+
     }
 
     func reloadData() {
         if data.status != .Success {
             print("Err: \(data.statusMessage)", terminator: "\n")
         } else {
-            self.tableView.reloadData()
+            //ensure reloadData call is in main queue
```
garvankeeley: Curious: do you have any reason to believe this code is not called from the main thread?
I mention because UI code isn't typically wrapped for main-thread safety.

Client/Frontend/Widgets/SiteTableViewController.swift
Hoid: I agree with @garvankeeley here. I don't think it makes sense to intentionally design a scenario where, if a bug happens, a crash occurs just so that it's easy to find/notice. We should be focused on the user's experience, not making our jobs easier, and crashes are never good for the user.

Client/Frontend/Widgets/SiteTableViewController.swift
diegodoval: reloadData in this class is called from a bunch of places (including when the profile object is set), and there's subclasses like BookmarksPanel, HistoryPanel and SearchViewController which end up calling reloadData directly or through that, as far as I can tell, but I haven't followed every path that leads there. Easier to wrap the update and be able to ignore what's upstream. If we're in the right thread, there's no context switching and the cost is near zero.

diegodoval:
This PR and branch no longer apply. Work was moved to another branch.

Issue #296, opened 2016-07-25
modify styling and dimensions of tabs in tabs bar


bradleyrichter:
Changes:

1. Increase tab height by roughly 4 pixels to match height of URL bar

2. reduce height of separator bar between tabs to match the appearance in the screen shot below (DD Addition - line should not be there to the left of the selected tab)

3. remove close buttons from non-selected tabs

![image](https://cloud.githubusercontent.com/assets/13509546/17111959/b6b011d6-5258-11e6-82cf-df374af2aa00.png)

garvankeeley:
First test of our QA Please label, and assigning to Alex. This will keep Alex informed of front-end changes, some changes might need additional hands-on testing (such as multi-device testing), but primarily this will keep him in-the-loop as the UI changes.

PR #314, opened 2016-08-02
Issue 296 followup

garvankeeley:
Follow commit for issue #296. 
My previous sloppy code made setting the tabs bar height a mess. The previous commit resized the tabs, but not the containing top toolbar.


brave/src/frontend/tabsbar/TabWidget.swift
```diff
@@ -123,7 +124,7 @@ class TabWidget : UIView {
             make.top.bottom.equalTo(self)
             make.left.lessThanOrEqualTo(close.snp_right)
             make.width.lessThanOrEqualTo(title.frame.width)
-            make.right.greaterThanOrEqualTo(self).inset(4)
+            make.right.greaterThanOrEqualTo(self).inset(labelInsetFromRight)
```
garvankeeley: @diegodoval It is possible to drag and drop the tabs, this function is part of the break/restore layout during this process. (The nsconstraint layout system does not like really aggressive collapsing and expanding of container views. Another place in the code that nsconstrants are a PITA, maybe would have been easier not to use them in retrospect.)

garvankeeley:
bar now looks like this:
screenshot 2016-08-01 21 08 31

PR #315, opened 2016-08-02
issue #307 iOS 8 screenshot crash: add extra guards on screenshot code (might stop crash?)

### LINKED ISSUE #307: fix screenshot code crash (iOS 8 only)
Crash is at drawViewHierarchyInRect within the function that draws the screenshot into a UIImage context.
	
UIViewExtensions.swift line 17
UIView.screenshot(CGFloat, offset : CGPoint?, quality : CGFloat) -> UIImage?
called from: ScreenshotHelper.swift line 41
                     ScreenshotHelper.(takeScreenshot(Browser) -> ()).(closure #1)

Screenshots are used mainly by the tab functionality to have a preview of the page available. See uses of ScreenshotHelper.takeScreenshot.

garvankeeley:
Issue #307. Not crashing in simulator, but drawViewHierarchyInRect has preconditions that size should be at least 1x1 and have a superview (according to others on StackOverflow who had similar crashes on older iOS versions)

### DIFF
--- a/Utils/Extensions/UIViewExtensions.swift
+++ b/Utils/Extensions/UIViewExtensions.swift
@@ -12,6 +12,10 @@ extension UIView {
     func screenshot(size: CGSize, offset: CGPoint? = nil, quality: CGFloat = 1) -> UIImage? {
         assert(0...1 ~= quality)
 
+        if size.width < 1 || size.height < 1 || superview == nil {
+            return nil
+        }
+
         let offset = offset ?? CGPointMake(0, 0)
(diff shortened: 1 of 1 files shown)


garvankeeley:
@diegodoval r?

I can see this code needing to be extra defensive, the calling code is a hack. Getting the screenshot of non-visible tabs requires temporarily adding the UIWebViews to the view hierarchy, let the event loop run so the view is properly added (I tried with no delay, no screenshot), then screenshot it, then remove it. 

Likely if I did a particular sequence in the UI I could get this to crash.

garvankeeley:
Landing, as it is no longer labeled as a fix, just some extra guard conditions. Original bug is still open and needs more research.

PR #316, opened 2016-08-02
Remove WKNavigationDelegate, which required fake WK* webkit objects.

garvankeeley:
This was originally done as a hack to quickly connect UIWebView code to use Firefox iOS interfaces without WKWebView. To conform to the interfaces in Fx iOS code, certain WKWebView API objects are needed, so I had faked them.

For one there is a minor memory leak reported with constructing parentless WKNavigation objects (actually, one is NEVER supposed to do this, but I did it anyway because of a time crunch).
Also, the code is difficult to follow with the fake objects being passed around (not to mention how confusing these fake objects are when reading the code).

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
@@ -3,8 +3,9 @@
 import Shared
 private let log = Logger.browserLogger
 
-extension BrowserViewController: WKNavigationDelegate {
-    func webView(webView: WKWebView, didStartProvisionalNavigation _: WKNavigation!) {
+extension BrowserViewController: WKCompatNavigationDelegate {
+
+    func webViewDidStartProvisionalNavigation(webView: UIWebView, url: NSURL?) {
         #if !BRAVE
             if tabManager.selectedTab?.webView !== webView {
                 return
@@ -18,7 +19,7 @@ extension BrowserViewController: WKNavigationDelegate {
         // If we are going to navigate to a new page, hide the reader mode button. Unless we
         // are going to a about:reader page. Then we keep it on screen: it will change status
         // (orange color) as soon as the page has loaded.
-        if let url = webView.URL {
+        if let url = tabManager.tabForWebView(webView)?.url {
             if !ReaderModeUtils.isReaderModeURL(url) {
                 urlBar.updateReaderModeState(ReaderModeState.Unavailable)
                 hideReaderModeBar(animated: false)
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -141,56 +135,17 @@ extension BrowserViewController: WKNavigationDelegate {
         // every time. There is no way around this prompt. (TODO Confirm this is true by adding them to the Info.plist)
 
         UIApplication.sharedApplication().openURL(url)
-        decisionHandler(WKNavigationActionPolicy.Cancel)
+        shouldLoad = false
     }
 
-    //    func webView(webView: WKWebView, didReceiveAuthenticationChallenge challenge: NSURLAuthenticationChallenge, completionHandler: (NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Void) {
```
garvankeeley: This for WKWebView basic auth, I was hoping to find some use for this for UIWebView auth but they are unrelated.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKUIDelegate.swift
```diff
@@ -125,16 +125,4 @@ extension BrowserViewController: WKUIDelegate {
 
         return false
     }
-
-    func webView(webView: WKWebView, decidePolicyForNavigationResponse navigationResponse: WKNavigationResponse, decisionHandler: (WKNavigationResponsePolicy) -> Void) {
```
garvankeeley: I never hooked up this interface, there is no equivalent notification in UIWebView

Client/Frontend/Browser/TabManager.swift
```diff
@@ -822,11 +822,11 @@ extension TabManager : WKNavigationDelegate {
     /// Called when the WKWebView's content process has gone away. If this happens for the currently selected tab
     /// then we immediately reload it.
 
-    func webViewWebContentProcessDidTerminate(webView: WKWebView) {
-        if let browser = selectedTab where browser.webView == webView {
-            webView.reload()
-        }
-    }
+//    func webViewWebContentProcessDidTerminate(webView: WKWebView) {
```
garvankeeley: hmm, should just remove, not comment out

Client/Frontend/Browser/TabManager.swift
```diff
@@ -843,92 +843,97 @@ extension TabManager {
...
-            delegate.webView?(webView, didCommitNavigation: navigation)
-        }
+    func insert(delegate: WKCompatNavigationDelegate) {
+        navDelegates.append(Weak_WKCompatNavigationDelegate(value: delegate))
     }
 
-    func webView(webView: WKWebView, didFailNavigation navigation: WKNavigation!, withError error: NSError) {
-        for delegate in delegates {
-            delegate.webView?(webView, didFailNavigation: navigation, withError: error)
+//    func webView(webView: WKWebView, didCommitNavigation navigation: WKNavigation!) {
```
garvankeeley: Ugh, I see I just commented out a bunch of functions. I was leaving them because at some point we will be back on WKWebView. Should I just delete these?

Client/Frontend/Reader/ReadabilityService.swift
```diff
@@ -41,7 +41,7 @@ class ReadabilityOperation: NSOperation, WKNavigationDelegate, ReadabilityBrowse
             let configuration = WKWebViewConfiguration()
             self.browser = Browser(configuration: configuration)
             self.browser.createWebview()
-            self.browser.navigationDelegate = self
+           /// self.browser.navigationDelegate = self
```
garvankeeley: I'll add a note to this before landing. I commented this out until we do reader mode, then this goes back on, and reader mode will be rewritten to support the new navigationDelegate class I created in this PR

garvankeeley:
@diegodoval r?

PR #317, opened 2016-08-02
Moving a db call off main thread

garvankeeley:
I am trying to catch any places th DB is called sync

We have a debug-only assertion to help catch these, this is the last one i can find.


Client/Frontend/Browser/LoginsHelper.swift
```diff
@@ -169,7 +169,9 @@ class LoginsHelper: BrowserHelper {
                 SnackButton(title: Strings.LoginsHelperSaveLoginButtonTitle, accessibilityIdentifier: "SaveLoginPrompt.saveLoginButton", callback: { (bar: SnackBar) -> Void in
                     self.browser?.removeSnackbar(bar)
                     self.snackBar = nil
-                    self.profile.logins.addLogin(login)
+                    succeed().upon { _ in // move off main thread
```
garvankeeley: All DB calls are on the queue created by the deferred classes, so an empty deferred is a handy way to move off to that thread. This is a Firefox iOS code convention, but it *is* odd-looking

PR #328, opened 2016-08-05
adding a default issue template

alexwykoff:
Added a default issue template to organize and speed up bug reports. It follows the style of the laptop issue reports.


garvankeeley:
lgtm!

PR #332, opened 2016-08-05
fix #329 isTopFrameRequest was incorrectly including the scheme

### LINKED ISSUE #329: Open in Background does not respect shield settings
Yes

**Describe the issue you encountered:**
While testing the HTTPS Everywhere functionality, I visited http://www.apple.com/ by opening the context menu and selecting "Open in Background". When visiting the site, I noticed the protocol remained http instead of the default of HTTPS Everywhere being enabled.

**Expected behavior:**
The browser should use whatever shield settings are enabled.

iPhone5

- Brave Version:
1.2.12 beta 1
- Steps to reproduce:
    1. Open a page which has a link to http://www.apple.com .
    2. Use the context menu to open the link in the background.
    3. Wait 30 seconds then switch to the newly created tab, check the protocol used.

- Screenshot if needed:

- Any related issues:

garvankeeley:
… in the comparison of requested url and main document url

### DIFF
--- a/brave/src/webview/BraveWebView.swift
+++ b/brave/src/webview/BraveWebView.swift
@@ -510,7 +510,8 @@ class BraveWebView: UIWebView {
     }
 
     class func isTopFrameRequest(request:NSURLRequest) -> Bool {
-        return request.URL == request.mainDocumentURL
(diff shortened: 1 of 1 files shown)


garvankeeley:
@diego r?

PR #335, opened 2016-08-07
Httpse performance: in mem cache of 1st db query

garvankeeley:
HTTPSE has 2 database queries, only the 2nd was cached, this adds a cache on the 1st query also. 
Without this fix we see a 15% performance hit:
screenshot 2016-08-07 15 53 13

With this fix we can see the off/on times are about the same:
screenshot 2016-08-07 15 48 11

### DIFF
--- a/brave/src/webfilters/HttpsEverywhere.swift
+++ b/brave/src/webfilters/HttpsEverywhere.swift
@@ -11,6 +11,7 @@ class HttpsEverywhere {
     var isNSPrefEnabled = true
     var db: Connection?
     let fifoCacheOfRedirects = FifoDict()
+    let fifoCacheOfDomainToIds = FifoDict()
 
     lazy var networkFileLoader: NetworkDataFileLoader = {
         let targetsDataUrl = NSURL(string: "https://s3.amazonaws.com/https-everywhere-data/\(dataVersion)/httpse.sqlite")!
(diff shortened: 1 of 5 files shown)


garvankeeley:
@diegodoval r?

brave/src/webfilters/HttpsEverywhere.swift
```diff
@@ -141,6 +142,10 @@ class HttpsEverywhere {
 
         var result = [Int]()
```
garvankeeley: this line should be moved into...

brave/src/webfilters/HttpsEverywhere.swift
```diff
@@ -141,6 +142,10 @@ class HttpsEverywhere {
...
 
+        if let cached = fifoCacheOfDomainToIds.getItem(domain) as? [Int] {
+            return cached
+        }
+
         if let row = db.prepare(query).generate().next() {
```
garvankeeley: var result = [Int]() goes here

brave/src/webfilters/HttpsEverywhere.swift
```diff
@@ -140,12 +140,15 @@ class HttpsEverywhere {
...
-        var result = [Int]()
-
         if let cached = fifoCacheOfDomainToIds.getItem(domain) as? [Int] {
             return cached
         }
 
+        var result = [Int]()
+        defer {
+            fifoCacheOfDomainToIds.addItem(domain, value: result)
+        }
```
garvankeeley: @diegodoval your defer sir!

Issue #308, opened 2016-07-27
crash/hang on removing tab 


diegodoval:
Crash is at 

            let next = self.neighborTab(index + 1)
of TabsBarViewController

TabsBarViewController.swift line 205
TabsBarViewController.(removeTab(TabWidget) -> ()).(closure #2)

probably something to do with next position not existing, but subsequent code presumes variable could be nil. Needs repro in simulator.

garvankeeley:
Full report:
```
Crashed: com.apple.main-thread
0  Client                         0x10009f47c TabsBarViewController.(removeTab(TabWidget) -> ()).(closure #2) (TabsBarViewController.swift:205)
1  UIKit                          0x1877fc12c -[UIViewAnimationBlockDelegate _didEndBlockAnimation:finished:context:] + 628
2  UIKit                          0x1877fbc28 -[UIViewAnimationState sendDelegateAnimationDidStop:finished:] + 312
3  UIKit                          0x1877fbab0 -[UIViewAnimationState animationDidStop:finished:] + 108
4  QuartzCore                     0x185155998 CA::Layer::run_animation_callbacks(void*) + 284
5  libdispatch.dylib              0x1820ad47c _dispatch_client_callout + 16
6  libdispatch.dylib              0x1820b2b84 _dispatch_main_queue_callback_4CF + 1844
7  CoreFoundation                 0x182618d50 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12
8  CoreFoundation                 0x182616bb8 __CFRunLoopRun + 1628
9  CoreFoundation                 0x182540c50 CFRunLoopRunSpecific + 384
10 GraphicsServices               0x183e28088 GSEventRunModal + 180
11 UIKit                          0x18782a088 UIApplicationMain + 204
12 Client                         0x100106074 main (main.swift:26)
```

garvankeeley:
In iTunes crash report, the report is in the line after that, in the SnapKit closure, which makes more sense.  The 'neighborTab' function is designed for nil handling, so it shouldn't crash:
```
    func neighborTab(i: Int) -> TabWidget? {
        return 0 ..< self.tabs.count ~= i ? self.tabs[i] : nil
    }
```

garvankeeley:
I wonder if the crash is triggered from deleting a tab in the Tabs Tray view vs. tapping the 'X' on a tab in the tabs bar.
[...]

PR #337, opened 2016-08-08
issue #308: tab removal crash

diegodoval:
Refactoring and changes in Tab adding/removal behavior to address #308 and potentially other tab problems related to it. We are losing one minor detail of the animation when removing (the tab removed doesn't compress down to 0 width).

Primarily, concentrated code for updates into a single method and ensured that changes to the structure happen before requesting a relayout/redraw.

### DIFF
--- a/brave/src/frontend/tabsbar/TabsBarViewController.swift
+++ b/brave/src/frontend/tabsbar/TabsBarViewController.swift
@@ -215,11 +215,11 @@ class TabsBarViewController: UIViewController {
         assert(index < self.tabs.count)
         
         if !self.isVisible {
-            self._removeTab(tab, atIndex: index)
+            _removeTab(tab, atIndex: index)
         }
         else {
             UIView.animateWithDuration(0.2, animations: {
-                self._removeTab(tab, atIndex: index)
+                _removeTab(tab, atIndex: index)
             })
         }
     }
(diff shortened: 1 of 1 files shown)


brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -181,40 +190,38 @@ class TabsBarViewController: UIViewController {
...
-            self.overflowIndicators()
+            self.recalculateTabView()
+        }
+        
+        assert(index < self.tabs.count)
+        
+        if !self.isVisible {
+            self._removeTab(tab, atIndex: index)
+        }
+        else {
```
garvankeeley: hate to nit, but for future: } else {

brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -181,40 +190,38 @@ class TabsBarViewController: UIViewController {
...
             })
-
-            tab.spacerRight.removeFromSuperview()
-            tab.removeFromSuperview()
             self.tabs.removeAtIndex(index)
-            self.overflowIndicators()
+            self.recalculateTabView()
+        }
+        
+        assert(index < self.tabs.count)
```
garvankeeley: can remove,  see line 194

brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -181,40 +190,38 @@ class TabsBarViewController: UIViewController {
...
+        
+        if !self.isVisible {
+            self._removeTab(tab, atIndex: index)
```
garvankeeley: _removeTab is func scope i think, checking...

PR #339, opened 2016-08-09
New tab anim is conflicting a bit with 'scroll to show selected tab' animation

garvankeeley:


brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -168,21 +171,24 @@ class TabsBarViewController: UIViewController {
...
+                self.recalculateTabView()
+                let w = self.calcTabWidth(self.tabs.count)
+                if self.tabs.count > 2 {
+                    self.scrollView.contentOffset = CGPoint(x: w * CGFloat(self.tabs.count - 2), y: 0)
                 }
-            
-        }
-        else {
+            }) { _ in
+                UIView.animateWithDuration(0.1) {
```
garvankeeley: For some reason, this is taking longer than 0.3 (0.2+0.1) to set tab alpha to 1, however looks good enough that we can move on

Issue #310, opened 2016-07-31
HTTPSE is the heaviest operation on page load


garvankeeley:
Other than UIWebView internals, HTTPSE sql db lookup is our heaviest operation. 
I ran a quick profile and found HTTPSE db lookup is clearly top-of-the-heap. 
TODO for me: put in a good capture from the profiler, I think if I load an http site with lots of subresource loading (requiring lots of Https-E candidate lookups) I can demonstrate a 1sec added page load time.

garvankeeley:
Added a fix for repeated loads here, HTTPSE is no-cost when hitting sites seen already in the current session: 
https://github.com/brave/browser-ios/pull/335

garvankeeley:
@diracdeltas I still haven't got a nice profile for this, I will do, but heads up that hitting the SQL db on first load of a site is adding 15-20% to load time. (in-mem caching of HTTPSE lookups in iOS app ensures that *repeating* page loads are pretty much never hitting the DB)

Not sure if the SQL queries themselves could be optimized, or if we can have some bot that browses the Alexa top 1000, and creates a bloom filter from that, and we can check the in-mem bloom filter before hitting the DB.

garvankeeley:
@alexwykoff this is an interesting one to watch out for. 

In the next beta, I will have https://github.com/brave/browser-ios/pull/335 landed, at that point, the bug will only be visible the first time you visit a page (until app restart). On fresh app start, when you load HTTP (non HTTPS) sites with lots of resources (pages like cnn.com make ~200 requests during load), each one of those requests gets checked in the HTTPSE database for upgrading from http to https. The SQL DB for this is >10MB and the queries are fuzzy comparisons IIRC (not simple 1 to 1 key lookups), and on mobile this adds up to a performance headache.

garvankeeley:
Another option is to regress this feature to only try HTTPSE lookup on the top-level doc. But I am rather attached to the coolness of HTTPSE upgrades of the resource loads.

diracdeltas:
the HTTPS Everywhere rules are also available as JSON on AWS but i'm guessing that won't be any faster. we added a LRU cache in desktop-browser for recently-used rulesets, maybe that would help here?

garvankeeley:
We have in-memory caches, so the hit is when visiting a site for the first time in an app session.

garvankeeley:
[...]

PR #346, opened 2016-08-10
HTTPSE: use cpp code from laptop/android.

garvankeeley:
- use alternate JSON parser, original code uses chromium lib
- use c++11 regex, original code uses 3rd party lib

Original: chromium/chrome/browser/net/blockers/blockers_worker.cc

Part of https://github.com/brave/browser-ios/issues/310

### DIFF
--- a/brave/src/webfilters/HttpsEverywhereCpp.mm
+++ b/brave/src/webfilters/HttpsEverywhereCpp.mm
@@ -1,0 +1,232 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+#include "HttpsEverywhereCpp.h"
+
+#include <sqlite3.h>
+#include <string>
+#include <vector>
+#include <sstream>
+#include <regex>
+#include "JsonCpp.h"
+
+using json = nlohmann::json;
+
+sqlite3 *httpse_db_ = 0;
+std::string dbFilePath;
+
+int SQLITEIdsCallback(void *a_param, int argc, char **argv, char **column) {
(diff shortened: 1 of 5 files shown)


garvankeeley:
@SergeyZhukovsky r?

garvankeeley:
@diegodoval not sure I'll land this on master until we have done more performance work. 
I removed all the caches that I have in the swift version, I can add in cpp caches, but we have bigger problems to deal with first

brave/src/webfilters/HttpsEverywhereCpp.mm
```diff
@@ -0,0 +1,232 @@
...
+        domain_to_check.insert(0, ".");
+        domain_to_check.insert(0, domains[i]);
+        std::string prefix;
+        if (i > 0) {
+            prefix = "*.";
+        }
+        query += prefix + domain_to_check + "'";
+    }
+
+    sqlite3_profile(httpse_db_, xProfile, NULL);
```
garvankeeley: needs to be off before landing

diegodoval:
change itself looks ok -- not sure this is the way to go though.

garvankeeley:
concerns ?

brave/src/webfilters/HttpsEverywhereCpp.mm
```diff
@@ -0,0 +1,232 @@
...
+    if (err != SQLITE_OK) {
+        // std::cout << "sqlite db open error " << dbFilePath.c_str() << ", err == " << err;
+        return false;
+    }
+
+    return true;
+}
+
+std::string applyHTTPSRule(const std::string& originalUrl, const std::string& rule) {
+    auto json_object = json::parse(rule);
```
garvankeeley: @bridiver applyHTTPSRule shows how the JSON is used

PR #361, opened 2016-08-16
Fix #205: Switch in settings for always on private browsing mode

### LINKED ISSUE #205: 'Always load in private mode' feature
I think this would appeal to some, for those who don't want to switch from Safari, they might use Brave as a private-mode-only browser.

garvankeeley:
Addresses this feature request.

### DIFF
--- a/Client/Application/QuickActions.swift
+++ b/Client/Application/QuickActions.swift
@@ -116,7 +116,7 @@ class QuickActions: NSObject {
 
     private func handleOpenURL(withBrowserViewController bvc: BrowserViewController, urlToOpen: NSURL) {
         // open bookmark in a non-private browsing tab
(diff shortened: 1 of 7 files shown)


Client/Application/QuickActions.swift
```diff
@@ -116,7 +116,7 @@ class QuickActions: NSObject {
 
     private func handleOpenURL(withBrowserViewController bvc: BrowserViewController, urlToOpen: NSURL) {
         // open bookmark in a non-private browsing tab
-        bvc.switchToPrivacyMode(isPrivate: false)
+        //bvc.switchToPrivacyMode(isPrivate: false)
```
garvankeeley: we aren't using quick actions yet, maybe could hook this up soon

PR #381, opened 2016-08-26
Remove project generation, use committed Xcode proj

garvankeeley:


garvankeeley:
@diegodoval can you try this? 
If you apply this as a patch, you will have to delete Client.xcodeproj and brave/BraveInfo.plist first.

garvankeeley:
Tested on another mac and LGTM

Issue #374, opened 2016-08-23
When shields are down, the individual controls should mirror the down status.


alexwykoff:
Description:
I toggled shields down. The individual controls still appear to be on. This is a mixed signal for the user.

**Did you search for similar issues before submitting this one?**
Yes

**Expected behavior:**
If I toggle shields down, the individual controls should also be reflect the down status.

- Device (iPhone5, iPhone6s plus, iPad 3, ?):
Tested on iPhone6s plus

- Brave Version:
1.2.12

- Steps to reproduce:
    1. Toggle shields down

- Screenshot if needed:
![img_0261](https://cloud.githubusercontent.com/assets/1434040/17879303/9b10778a-68b6-11e6-8ccf-e8c55a9e6add.PNG)

- Any related issues:

garvankeeley:
Should they become disabled (and change color to indicate their disabled state)?

alexwykoff:
@bradleyrichter should have the call on that. My intuition leads me to believe that your suggestion to change color is probably the most appropriate.

bradleyrichter:
If the disabled switches are still showing as on, then the user must question if they are disabled (forced) on? (on but not selectable)

I think it is clearest to move the switch to off and disable it so that there is no confusion. (other than temporarily losing reference to it's intended state when shields are up.)

garvankeeley:
Fixed by #392

PR #391, opened 2016-09-01
When shields are down, the individual controls should mirror the down…

jamesmudgett:
UI fix for issue #374


garvankeeley:
Nice! can you remove the whitespace changes? It makes the blame view so much cleaner when only lines that are changed are specific to the fix. (It is wierd all the blank lines got changed, maybe Xcode went and fixed all the formatting on the blank lines.)

garvankeeley:
Closing, was resubmitted with nits fixed as PR #392

PR #392, opened 2016-09-01
When shields are down, the individual controls should mirror the down…

jamesmudgett:
… status. #374


garvankeeley:
Thanks James!

PR #413, opened 2016-09-08
Bookmark edit fixes

diegodoval:
This set of changes fixes the various issues associated with editing by using a simpler/more 'standard' interface, ie., copying Safari's in-table form for editing bookmarks.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -9,4 +9,5 @@ github "drmohundro/SWXMLHash"       "2.1.0"
 github "DaveWoodCom/XCGLogger"      "Version_3.5.1"        # Previously: swift_1.2
 github "fluffyemily/SwiftKeychainWrapper" "extension-compatibility"					# default to latest tag
 github "mixpanel/mixpanel-iphone"  "v3.0.3"
+github "xmartlabs/Eureka" "1.7.0"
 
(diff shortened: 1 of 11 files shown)


garvankeeley:
please remove extra project file changes from commit 5f80c094eaed26ae9de8f6f156eb47d07dbb8321

garvankeeley:
Label overlapping on entering edit mode is still not working too well. It briefly overlaps before snapping to the correct size, and it doesn't animate the size change. We can land without the animation, but no with the overlapping. Here is a POC  patch that fixes the bug by hacking the root problem of TwoLineCell.layoutSubviews() . However, this will break history panel cells, because the code is shared. 

https://gist.github.com/garvankeeley/07a91aa7107440e423555639a47b2fa7

garvankeeley:
Other problem is the shadowing on the header during navigation, we should fix that before landing.

diegodoval:
Other than the brief shadow on the header I don't see problems left, can we confirm?

garvankeeley:
Still seeing blank folder name when I create new folder, until table refresh

garvankeeley:
I'll rebase this branch and land it.

There is still an ugly shadow header navigation bug left that needs to be fixed, but that is getting filed

PR #422, opened 2016-09-12
Improvements for 'Settings > Report a Bug'

### LINKED ISSUE #331: Improvement for Settings > Report a Bug
(Related to #330) -- include this information within the email that is sent for reporting a bug.

Include the information at the end, format (note that in the format the square brackets are required to have a boundary around the version string that can be efficiently processed later):

```
----------------------------------------
Please do not modify the information that follows. 
It is completely anonymous and necessary for the Brave Team 
to evaluate reports, issues and bugs. Thanks!
----------------------------------------
iOS version:		[IOS_VERSION]
Device:				[DEVICE]
Brave version:		[BRAVE_VERSION]
----------------------------------------
```

### LINKED ISSUE #406: 'report a bug' (in settings) no response to button if there's no email account configured
On a device with no email setup (easiest is to try in simulator) 'report a bug' just seems to freeze. Need to show an error message.

diegodoval:
Fixed #406 - ‘report a bug' (in settings) no response to button if there's no email account configured

Fixed #331 - Improvement for Settings > Report a Bug

### DIFF
--- a/brave/src/frontend/BraveSettingsView.swift
+++ b/brave/src/frontend/BraveSettingsView.swift
@@ -309,7 +309,6 @@ class BraveSupportLinkSetting: Setting, MFMailComposeViewControllerDelegate {
     }
 
     override func onClick(navigationController: UINavigationController?) {
-        
         //MFMailComposeViewController is wonky on simulator, detect & avoid
         let isDevice = (TARGET_OS_SIMULATOR == 0)
         
@@ -325,14 +324,18 @@ class BraveSupportLinkSetting: Setting, MFMailComposeViewControllerDelegate {
             mailComposerVC.setToRecipients([reportBugMailAddress])
             mailComposerVC.setSubject(reportBugMailSubject)
             
+            let iOSVersionLabel = NSLocalizedString("iOS version", comment: "iOS version")
+            let deviceLabel = NSLocalizedString("Device", comment: "Device")
+            let braveVersionLabel = NSLocalizedString("Brave version", comment: "brave version")
+            
             let appVersionString: String = NSBundle.mainBundle().objectForInfoDictionaryKey("CFBundleShortVersionString") as! String
             let buildNumber: String = NSBundle.mainBundle().objectForInfoDictionaryKey("CFBundleVersion") as! String
             let versionAndBuildNumber: String = "\(appVersionString) (\(buildNumber))"
             
             var message = reportBugMailBody
-            message += "iOS version: [\(UIDevice.currentDevice().systemName)/\(UIDevice.currentDevice().systemVersion)]\n"
-            message += "Device: [\(UIDevice.currentDevice().name)/\(UIDevice.currentDevice().model)]\n"
-            message += "Brave version: [\(versionAndBuildNumber)]\n"
+            message += "\(iOSVersionLabel) [\(UIDevice.currentDevice().systemName)/\(UIDevice.currentDevice().systemVersion)]\n"
+            message += "\(deviceLabel): [\(UIDevice.currentDevice().name)/\(UIDevice.currentDevice().model)]\n"
+            message += "\(braveVersionLabel): [\(versionAndBuildNumber)]\n"
             message += "-----------------------------\n"
             mailComposerVC.setMessageBody(message, isHTML: false)
             
--- a/brave/src/frontend/BraveSettingsView.swift
+++ b/brave/src/frontend/BraveSettingsView.swift
@@ -4,6 +4,7 @@
     import Crashlytics
 #endif
 import Shared
+import MessageUI
 
 let kPrefKeyNoScriptOn = "noscript_on"
 let kPrefKeyFingerprintProtection = "fingerprintprotection_on"
@@ -126,7 +127,7 @@ class BraveSettingsView : AppSettingsTableViewController {
         settings += [
             SettingSection(title: NSAttributedString(string: NSLocalizedString("Support", comment: "Support section title")), children: [
                 ShowIntroductionSetting(settings: self),
-                BraveSupportLinkSetting(),
+                BraveSupportLinkSetting(parentVC: self),
                 BravePrivacyPolicySetting(), BraveTermsOfUseSetting(),
                 ])]
         //#endif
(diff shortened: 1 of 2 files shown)


brave/src/frontend/BraveSettingsView.swift
```diff
@@ -277,14 +278,73 @@ class PasswordsClearable: Clearable {
...
+            mailComposerVC.setMessageBody(message, isHTML: false)
+            
+            parentViewController.presentViewController(mailComposerVC, animated: true, completion: nil)
+        } else {
+            showMailErrorDialog()
+        }
+    }
+    
+    func showMailErrorDialog() {
+        let sendMailErrorAlert = UIAlertView(title: "Error sending email", message: "It appears you're not setup for email on this device.  Please check your email configuration and try again.", delegate: self, cancelButtonTitle: "OK")
```
garvankeeley: NSLocalizedString

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -277,14 +278,73 @@ class PasswordsClearable: Clearable {
...
-
+    
+    func mailComposeController(controller: MFMailComposeViewController, didFinishWithResult result: MFMailComposeResult, error: NSError?) {
+        
+        controller.dismissViewControllerAnimated(true) {
+            if result == MFMailComposeResultSaved || result == MFMailComposeResultSent {
+                postAsyncToMain {
+                    //only show 'thanks' dialog when Saved/Sent. 
+                    //TODO turn this into a Toast-type notification that goes away by itself.
+                    let sendFeedbackThanks = UIAlertView(title: "Thanks!", message: "Thank you for your feedback!", delegate: self, cancelButtonTitle: "OK")
```
garvankeeley: NSLocalizedString

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -277,14 +278,73 @@ class PasswordsClearable: Clearable {
...
     override var title: NSAttributedString? {
         return NSAttributedString(string: NSLocalizedString("Report a bug", comment: "Show mail composer to report a bug."), attributes: [NSForegroundColorAttributeName: UIConstants.TableViewRowTextColor])
     }
-
+    
+    func mailComposeController(controller: MFMailComposeViewController, didFinishWithResult result: MFMailComposeResult, error: NSError?) {
+        
+        controller.dismissViewControllerAnimated(true) {
+            if result == MFMailComposeResultSaved || result == MFMailComposeResultSent {
+                postAsyncToMain {
```
garvankeeley: Code comment why this is needed

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -277,14 +278,73 @@ class PasswordsClearable: Clearable {
...
+                    //only show 'thanks' dialog when Saved/Sent. 
+                    //TODO turn this into a Toast-type notification that goes away by itself.
+                    let sendFeedbackThanks = UIAlertView(title: "Thanks!", message: "Thank you for your feedback!", delegate: self, cancelButtonTitle: "OK")
+                    sendFeedbackThanks.show()
+                }
+            }
+            else if result == MFMailComposeResultFailed {
```
garvankeeley: nit: convention is `} else {`

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -277,14 +278,73 @@ class PasswordsClearable: Clearable {
...
+        //MFMailComposeViewController is wonky on simulator, detect & avoid
+        let isDevice = (TARGET_OS_SIMULATOR == 0)
+        
+        if MFMailComposeViewController.canSendMail() && isDevice {
+            
+            let reportBugMailAddress = "support+ios@brave.com"
+            let reportBugMailSubject = NSLocalizedString("Brave for iOS Feedback", comment: "email subject")
+            let reportBugMailBody = "\n\n\n--------------------\nPlease do not modify the information that follows.\nIt is completely anonymous and necessary for the Brave Team to evaluate reports, issues and bugs. Thanks!\n-----------------------------\n";
+
+
```
garvankeeley: nit: \n

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -277,14 +278,73 @@ class PasswordsClearable: Clearable {
...
-        UIApplication.sharedApplication().openURL(NSURL(string: "mailto:support+ios@brave.com")!)
+        
+        //MFMailComposeViewController is wonky on simulator, detect & avoid
+        let isDevice = (TARGET_OS_SIMULATOR == 0)
+        
+        if MFMailComposeViewController.canSendMail() && isDevice {
+            
+            let reportBugMailAddress = "support+ios@brave.com"
+            let reportBugMailSubject = NSLocalizedString("Brave for iOS Feedback", comment: "email subject")
+            let reportBugMailBody = "\n\n\n--------------------\nPlease do not modify the information that follows.\nIt is completely anonymous and necessary for the Brave Team to evaluate reports, issues and bugs. Thanks!\n-----------------------------\n";
```
garvankeeley: Quite verbose, can we shorten this? Perhaps
"If your report is a bug, the following version information will help us diagnose it."

brave/src/frontend/BraveSettingsView.swift
diegodoval: Why is what needed specifically?

garvankeeley: this code is main thread, so the reason isn't clear

diegodoval: In this particular case we're in the event thread but we're also receiving events -- I think it's good practice in general to ensure that we're pushing something asynchronously and letting the event processing continue its course.

brave/src/frontend/BraveSettingsView.swift
diegodoval: changed it to:

```
--
App & Device Version Information:
```

brave/src/frontend/BraveSettingsView.swift
garvankeeley: Assuming it is necessary in a particular case, isn't it better to deal with a specific case? If there is a specific area of UIKit that is sync and problematic, then I say yes to this –but the calls here are async already, like show() for a dialog. (I can't think of a case where this is needed, particularly in this codebase as underlying heavy sync calls are wrapped in Deferred). Admittedly, async dispatch no main is used sometimes to work around buggy/wierd UIKit behaviour where the UI needs to update before doing a consecutive UI step.

If it is a good practice, then note, it isn't done anywhere else in Brave or Firefox iOS. If we decided this was a standard, we would have hundreds of places in the code to change. If this a practice, it is one that must be rare, I haven't seen this used before and I assume other devs will be the same, so the bar is high to get buy-in on doing this.

PR #425, opened 2016-09-13
Fix #423: show better default set of sites and require more site visits to bump defaults

### LINKED ISSUE #423: Top Sites: show better default set and require more site visits to bump the defaults off
This is a repair task for the bad top site behaviour we have currently.
A site can be visited once (or a few times) and show up on top sites. 
1) we should make sure a site has been visited many times before bumping a default site off

2) we should have a better list of popular sites

garvankeeley:
fix #423
Change default sites to: google, yahoo, facebook, youtube, amazon, twitter, wikipedia, duckduckgo
Require 20 site visits before showing up on top sites


garvankeeley:
@jamesmudgett I already landed this so I could spit out a beta with it. The color scheme isn't great.  Perhaps not-even-close-to-not-great and shouldn't land in release like this?

PR #435, opened 2016-09-19
Telemetry hookup

garvankeeley:
cc @diegodoval


diegodoval:
It seems we are using the Objective-C API, any reason not to use their Swift library? https://mixpanel.com/help/reference/swift

diegodoval:
Telemetry calls look ok to me. Going to double check the initial flow now.

garvankeeley:
i am using mixpanel-iphone, i see i should be using https://github.com/mixpanel/mixpanel-swift

garvankeeley:
Landing as-is, mixpanel-swift has carthage problems documented (which I just ran into): https://github.com/mixpanel/mixpanel-swift/issues/18
We can update later when we dig deeper into mixpanel and this is resolved, for now it is just one call we are using.

Issue #428, opened 2016-09-17
Localization settings UI bug: UITableViewCells with detailTextLabel overflowing


garvankeeley:
This brutally difficult to fix. The cells with detailTextLabel (the label on the right) are not handling long strings. This is a screenshot of testing using XCode's 'double length pseudolanguage'. Both the textLabel and the detailTextLabel need to shrink, although simply fixing the textLabel to 2/3 screen max width and then shrinking the font would workaround nearly all cases:

![simulator screen shot sep 16 2016 10 33 15 pm](https://cloud.githubusercontent.com/assets/5495083/18605351/a8a2cc76-7c5d-11e6-9361-385c4be6f2f1.png)

garvankeeley:
@jamesmudgett help!

PR #459, opened 2016-09-28
Adjust fontsize to fit width, limit lines on detail label (Issue #428)

jamesmudgett:


garvankeeley:
lgtm! (haven't tried in xcode double-length string mode yet thought)

PR #467, opened 2016-09-30
Fix #436 - Folder creation

### LINKED ISSUE #436: Bookmarks: ugly flickering shadow when navigating folders
When a new view controller is pushed on the bookmark table (e.g. entering edit mode on a cell, or going into a subfolder) the drop shadow from the top view that appears over the panel seems to move along with the push animation, then disappear. 

Have tried various things to remove it, without success. Removing the animation altogether removes the problem but looks worse.

jamesmudgett:


garvankeeley:
I am not sure what bug this is for, not #436 though, that bug is a flickering shadow in the header when navigating

PR #489, opened 2016-10-11
Fix #446, remove over-influencing localVisit

### LINKED ISSUE #446: Top Sites: is based on page visits, not domain visits, see if this can be changed
I thought the top sites frecency query was based on a domain visit count, but it is a count of visiting a specific page, which isn't what we want. Regardless of having visited site.com/url1,  site.com/url2, or site.com/url3, we want site.com to appear on the top sites.

jamesmudgett:
This needs to be tested thoroughly, the idea is that localVisitCount shouldn't over influence frecency as it is currently doing that and causing some sites to appear in topsides, on few visits.

### DIFF
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -85,7 +85,7 @@ func getRemoteFrecencySQL() -> String {
 }
 
 func getLocalFrecencySQL() -> String {
-    let visitCountExpression = "((2 + localVisitCount) * (2 + localVisitCount))"
+    let visitCountExpression = "localVisitCount"
     let now = NSDate.nowMicroseconds()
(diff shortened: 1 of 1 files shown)


Storage/SQL/SQLiteHistory.swift
```diff
@@ -85,7 +85,7 @@ func getRemoteFrecencySQL() -> String {
 }
 
 func getLocalFrecencySQL() -> String {
-    let visitCountExpression = "((2 + localVisitCount) * (2 + localVisitCount))"
+    let visitCountExpression = "localVisitCount"
```
garvankeeley: Where does localVisitCount come from?
I can't recall the DB structure -and of course the lack of ORM-ness makes the code brutal to quickly see where fields are coming from. Is this related to the 'visits' table? (that table is basic IIRC just has a site and a date visited)

garvankeeley:
closing, as james and I have bigger fixing plans for this

PR #502, opened 2016-10-14
Fix #349, Remove snackbar for thridparty password managers, use as ke…

### LINKED ISSUE #349: Remove 3rd Party Password Snackbar
This was an experimental feature that has had its day in the sun.
It incomplete, and I am not convinced many are using it. 3rd party password manager users are familiar with pressing the Share button to fill in passwords.

@alexwykoff @diegodoval IIRC you both notice the lack of awesomeness of this feature.

Diego, I liked your suggestion about adding a button to keyboard top bar for PW manager access, when a login or PW field has been selected. I do find myself tapping the login field before realizing I now covered the Share button, have to tap away to close the keyboard, then select the Share button.

jamesmudgett:
…yboard accessory instead.

TODO: Still need to work on the utilization of the third party password manager extensions to avoid doing hacky share-sheet auto selection.


bradleyrichter:
thanks!

PR #517, opened 2016-10-24
Selecting password helper footer does not dismiss keyboard

jamesmudgett:


jamesmudgett:
Still need to work on getting Password extensions to display without requiring action sheet.

garvankeeley:
Thanks!

PR #519, opened 2016-10-24
Fix development team setting in xcode

garvankeeley:
Change to using a single local-def.xcconfig file. In this file, one can set their dev team.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,12 +74,10 @@ MozBuildID.xcconfig
 Carthage/Build
 
 node_modules/
-build-id.xcconfig
 ABPFilterParserData.dat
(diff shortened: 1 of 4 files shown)


jamesmudgett:
Nice work! Minus having to dig up the Team ID on my account process worked seamlessly, might be great to also include team id flag in setup.sh and have that write to file, to avoid additional steps.

garvankeeley:
Did you get entitlements error building to device? That should be fixed (well it seemed to have stopped complaining for me)

jamesmudgett:
No error building to device. No capabilities needed to be disabled. 👍

PR #528, opened 2016-10-27
JS Alert can be shown for non-active webview

garvankeeley:
this could be used for spoofing
Test case included.

@jamesmudgett


garvankeeley:
landing and seeing if iTunes upload accepts it

PR #529, opened 2016-10-28
Updated Password Manager flow for iPad

jamesmudgett:


garvankeeley:
Please review my commit, I tried to make it more readable, which was really the fault of my original code. I tried to make it really clear that the built-in password manager doesn't use this new button, as the built-in PW manager is an auto-fill-only pw manager.

brave/src/page-hooks/LoginsHelper+1PW.swift
```diff
@@ -162,7 +160,7 @@ extension LoginsHelper {
     @objc func onExecuteTapped(sender: UIButton) {
         self.browser?.webView?.endEditing(true)
 
-        let automaticallyPickPasswordShareItem = (ThirdPartyPasswordManagerSetting.currentSetting != nil) ? ThirdPartyPasswordManagerSetting.currentSetting! != ThirdPartyPasswordManagers.UseBuiltInInstead : false
+        let automaticallyPickPasswordShareItem = ThirdPartyPasswordManagerSetting.currentSetting != nil
```
garvankeeley: Maybe i should have commented, or assert(ThirdPartyPasswordManagerSetting.currentSetting != builtin), because that is a precondition of hitting this function

brave/src/page-hooks/LoginsHelper+1PW.swift
```diff
@@ -10,11 +10,11 @@ var noPopupOnSites: [String] = []
...
 
-
+typealias ThirdPartyPasswordManagerType = (displayName: String, cellLabel: String, prefId: Int)
 struct ThirdPartyPasswordManagers {
-    static let UseBuiltInInstead = (displayName: "Don't use", cellLabel: "", prefId: 0)
-    static let OnePassword = (displayName: "1Password", cellLabel: "1Password", prefId: 1)
-    static let LastPass = (displayName: "LastPass", cellLabel: "LastPass", prefId: 2)
+    static let UseBuiltInInstead:ThirdPartyPasswordManagerType = (displayName: "Don't use", cellLabel: "", prefId: 0)
+    static let OnePassword:ThirdPartyPasswordManagerType = (displayName: "1Password", cellLabel: "1Password", prefId: 1)
+    static let LastPass:ThirdPartyPasswordManagerType = (displayName: "LastPass", cellLabel: "LastPass", prefId: 2)
```
garvankeeley: I should have a note, when we add another one, stop and just classes and enums, no more tuple, it is making for ugly code

jamesmudgett:
Looks good

brave/src/page-hooks/LoginsHelper+1PW.swift
jamesmudgett: Did this break things?

garvankeeley: No, then yes, and now no again :), your inital PR was fine, I added my additional commit, which then broke it. I had the unset state of that wrong, which i just fixed 
https://github.com/brave/browser-ios/commit/f4ed6e839491f38247fb1e2e6b0b710fae1e7055

PR #530, opened 2016-10-31
Cleaning up logic for hide/show 1PW button (some of which I just broke)

garvankeeley:


Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1182,17 +1182,14 @@ extension BrowserViewController: KeyboardHelperDelegate {
...
                 if UIDevice.currentDevice().userInterfaceIdiom == .Pad {
                     self.urlBar.pwdMgrButton.hidden = !shouldShow
-                    
-                    guard let current = ThirdPartyPasswordManagerSetting.currentSetting else { return }
-                    if current == ThirdPartyPasswordManagers.OnePassword {
-                        self.urlBar.pwdMgrButton.setImage(UIImage(named: "passhelper_1pwd")?.imageWithRenderingMode(.AlwaysTemplate), forState: .Normal)
-                    } else if current == ThirdPartyPasswordManagers.LastPass {
-                        self.urlBar.pwdMgrButton.setImage(UIImage(named: "passhelper_lastpass")?.imageWithRenderingMode(.AlwaysTemplate), forState: .Normal)
-                    }
-                    self.urlBar.updateConstraints()
```
garvankeeley: AFAIK setNeedsUpdateConstraints() is required for the layout manager to trigger, calling updateConstraints directly may not have any effect if the layout manager doesn't also run a layout update.

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1182,17 +1182,14 @@ extension BrowserViewController: KeyboardHelperDelegate {
         }
 
         if let loginsHelper = tabManager.selectedTab?.getHelper(LoginsHelper) {
```
garvankeeley: I am seeing keyboardWillShowWithState getting called after a hide. Unless I am loosing my mind, which is always a distinct possibility :). Looking at this now. So bizarre.

brave/src/page-hooks/LoginsHelper+1PW.swift
```diff
@@ -34,7 +34,7 @@ extension LoginsHelper {
                 return // No 3rd party password manager installed
             }
 
-            postAsyncToMain(0.1) {
+            postAsyncToMain {
```
garvankeeley: just a cleanup item, not a fix for anything

garvankeeley:
I think we can put this to rest for while, seems to be working well now

PR #561, opened 2016-11-10
Fix issue with bookmarks sort interrupted gestures

jamesmudgett:


jamesmudgett:
There was an issue when editing bookmarks and trying to sort the rows, occasionally getting into this strange state where the drag and drop would be interrupted and not allow the drag.

garvankeeley:
Thanks for jumping on this one!

Issue #277, opened 2016-07-04
Fresh tab: add overview text area to brave panel when disabled


garvankeeley:
Tapping it does nothing, which is confusing

bradleyrichter:
Here is a suggested improvement where we use the opportunity to provide some helpful info for the user who chooses the brave button at the new-tab page.

![image](https://cloud.githubusercontent.com/assets/13509546/17154498/25b71d1c-5336-11e6-819c-9a3c1899b101.png)

bradleyrichter:
cc @diegodoval

garvankeeley:
Cool idea, and I don't recall and code structure/flow issue that would make this a problem to implement (might be easy)

garvankeeley:
fixed by https://github.com/brave/browser-ios/pull/599

PR #599, opened 2016-11-30
Issue #277, added shields overview to brave panel when disabled

ryan-liveware:
I added the shields overview to the brave panel when disabled. The panel uses a property named isShowingOverview and a function called isShowingHome() to determine the overview state. The state is based on the existence of the home panel controller. I used the existence of the home panel controller as the condition because that is what canShow was using.


garvankeeley:
Nicely done!

PR #601, opened 2016-12-03
Issue 600, right panel - separation of setting up the views, and updating them

jhreis:
The idea behind this PR is to separate the logic that used to initially construct the right panel, and the logic that needs to be adjusted dynamically (view size, button states, domain name label, and the new Shields Overview container's visibility).

On showing the right panel, constraints are now dynamically set. This keeps constraints clean, and prevents duplicate views from being added to the right-side panel.

I did notice there were lots of constraint errors on the logs (previously and currently present). These _seem_ to be failing correctly for _now_, but should probably be ironed out.

### DIFF
--- a/brave/src/frontend/sidepanels/BraveRightSidePanel.swift
+++ b/brave/src/frontend/sidepanels/BraveRightSidePanel.swift
@@ -463,9 +463,7 @@ class BraveRightSidePanelViewController : SidePanelBaseViewController {
     }
     
     func toggleInteractionEnabled(enabled: Bool) {
-        for toggle: UISwitch in views_toggles {
-            toggle.enabled = enabled && !isShowingOverview
-        }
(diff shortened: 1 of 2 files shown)


garvankeeley:
Yup, this is much better than the messiness that was there.

ryan-liveware:
After switching to landscape mode and back to portrait, the switches stop working in the brave panel. The scrollView contentSize is not being addressed in this PR. Also, the shields switch is off center when the overview is showing.

jhreis:
The shield switch is pinned to the bottom to have more universal constraints. This can easily be fixed by adjusting the containing view's frame to align with the switch. `setupContainerViewSize` is called on showing the panel which solves the contentSize issue, and I don't think rotation logic was changed at all. @ryan-liveware can you provide steps-to-reproduce for the switch bug? I cannot seem to reproduce it.

PR #604, opened 2016-12-08
Updated 'Report a bug' link to point to community.brave.com

alexwykoff:
Since community is live, it seems best to redirect users there for support.

Auditors: @garvankeely @jamesmudgett


garvankeeley:
Great stuff! I am going to open a new PR that has a different approach of opening discourse in a new tab

PR #617, opened 2016-12-15
Global Shield Stats on Top Sites

jamesmudgett:
Global Shield Stats - Display Brave shield stats on top sites (new tab) screen.

1) Multiple tabs stats may be updated by observing changes on Global Stats Singleton.
2) Slight variation from initial design(laptop) that contained a separation between ads and trackers blocked. Those are not independently tracked, so it has been combined int one stat "Ad/Trackers Blocked".
3) Shield stats are stored in NSUserDefaults, and only clear if app is deleted/reinstalled.
4) Color scheme reflects stats in Brave Shields Panel. 

There are likely several improvements to be made to this. Please review.

Thanks!


Client.xcodeproj/project.pbxproj
```diff
@@ -3165,44 +3171,54 @@
 				TargetAttributes = {
 					288A2D851AB8B3260023ABC3 = {
 						CreatedOnToolsVersion = 6.2;
+						DevelopmentTeam = 5D642W5D2P;
```
garvankeeley: Our project system results in local changes like this getting committed accidentally. I really should try to fix this. Can you unstage this part of the file with the DevTeam set? PITA I realize.

garvankeeley:
Tried it out, looking very cool!

PR #656, opened 2017-01-18
Add regional adblock

### LINKED ISSUE #564: Regional adblock
Adblock is currently using a data file that is not ideal for certain markets. We now have some locale-specific data files on desktop browser, look at hooking those up in mobile also.

garvankeeley:
Fix #564

@jhreis I need to research how to test this:
* without manually setting the locale in the simulator run scheme.
* check if a particular known ad is blocked for a specified region

Maybe I'll make that a separate follow up bug, or we can keep this open while test cases are worked out.

BTW, the pref for this is a but odd: it only defaults to on for well-tested regions, otherwise the user must manually toggle the pref on for their region.

### DIFF
--- a/brave/src/webfilters/AdBlocker.swift
+++ b/brave/src/webfilters/AdBlocker.swift
@@ -3,41 +3,100 @@
 import Foundation
 import Shared
 
-private let _singleton = AdBlocker()
+class AdblockNetworkDataFileLoader: NetworkDataFileLoader {
+    var lang = "en"
+}
+
+typealias localeCode = String
 
 class AdBlocker {
+    static let singleton = AdBlocker()
+
     static let prefKey = "braveBlockAdsAndTracking"
     static let prefKeyDefaultValue = true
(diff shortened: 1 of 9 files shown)


brave/src/webfilters/AdBlocker.swift
```diff
@@ -163,12 +226,24 @@ class AdBlocker {
             }
         }
 
-        let isBlocked = abpFilterLibWrapper.isBlockedConsideringType(url.absoluteString,
-                                                                     mainDocumentUrl: mainDocDomain,
-                                                                     acceptHTTPHeader:request.valueForHTTPHeaderField("Accept"))
+        var isBlocked = false
+        var blockedByLocale = ""
+        for (locale, adblocker) in abpFilterLibWrappers {
```
garvankeeley: Each abpFilterLibWrapper handles one adblock file. We can support multiple adblock files in future, but I only ever load 2 for now. @jhreis

PR #657, opened 2017-01-18
Safebrowsing v2

### LINKED ISSUE #646: safebrowsing upgrade to v2
@bbondy is there code that also needs to change? If I simply load V2 of this file (that is the only change I made):
https://s3.amazonaws.com/safe-browsing-data/2/SafeBrowsingData.dat
Then all sites I tried are coming back as positive hits (amazon, yahoo, etc).

garvankeeley:
fix #646 

Note the URL changed along with the version number

### DIFF
--- a/brave/src/webfilters/NetworkDataFileLoader.swift
+++ b/brave/src/webfilters/NetworkDataFileLoader.swift
@@ -89,7 +89,7 @@ class NetworkDataFileLoader {
             }
             else {
                 if let data = data, response = response as? NSHTTPURLResponse {
-                    if response.statusCode / 100 == 4 { // error
+                    if 400...499 ~= response.statusCode { // error
                         print("Failed to download, error: \(response.statusCode), URL:\(response.URL)")
                     } else {
                         let etag = response.allHeaderFields["Etag"] as? String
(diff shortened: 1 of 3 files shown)


garvankeeley:
hold on, adding a test case for this

brave/src/webfilters/NetworkDataFileLoader.swift
```diff
@@ -89,8 +89,12 @@ class NetworkDataFileLoader {
             }
             else {
                 if let data = data, response = response as? NSHTTPURLResponse {
-                    let etag = response.allHeaderFields["Etag"] as? String
-                    self.finishWritingToDisk(data, etag: etag)
+                    if response.statusCode / 100 == 4 { // error
```
jhreis: Any benefit of this over just checking `== 400`?

garvankeeley: Any 4xx series error is bad

jhreis: Ah, yes, can also use Swift's pattern matching `~=`:
`if 400...499 ~= response.statusCode { // error`

But I think either is fine.

garvankeeley: that looks better, changing!

Issue #705, opened 2017-02-08
Folder names cannot be edited to have space-only titles


jhreis:
Description:
New folders can be created with just spaces ' ' (expected behavior), but folders names cannot be edited with just spaces (input is rejected). The naming behavior/validation should be the same for creating and editing folder names.

**Did you search for similar issues before submitting this one?**
Yes

**Expected behavior:**
Editing folder name should allow space-only names

- Brave Version:
1.3

- Steps to reproduce:
    1. Create folder with name (e.g. 'test')
    2. Edit bookmarks
    3. Select newly created folder
    4. Attempt to edit the folder name to only include spaces
    5. Go back and see the name did not change

- Any related issues:
#513 - Not allowing 'empty' folder names (but allowing spaces).

garvankeeley:
Agreed with the space-only names, no reason a user can't do that if they don't feel like naming a folder.

PR #716, opened 2017-02-13
Fix: #705 Now folders can have space only titles

ViktorSimko:


jhreis:
May not be necessary to merge, but would be nice to gets some automated tests in for these folder titles at some point (creating and updating).

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -209,6 +218,9 @@ class BookmarkEditingViewController: FormViewController {
                 row.title = Strings.Name
                 row.value = bookmark.title
                 self.titleRow = row
+            }.cellSetup { cell, row in
+                cell.textField.addTarget(EurekaTextRowHelper.self, action: #selector(EurekaTextRowHelper.replaceNormalSpacesWithNonBreakingSpaces(_:)), forControlEvents: .EditingChanged)
+                cell.textField.addTarget(EurekaTextRowHelper.self, action: #selector(EurekaTextRowHelper.replaceNonBreakingSpacesWithNormalSpaces(_:)), forControlEvents: .EditingDidEnd)
```
garvankeeley: Hi I'm not familiar enough with the BM editing code. This means: during editing, a space gets replaced with an nbsp and then when completed gets converted back? Why is this needed? Thanks

ViktorSimko: Hi 

Yes. 
I think it's needed because spaces doesn't appear at the end of the text in right-aligned text fields.
And also when you try to rename a folder to a space only name and delete the old name, then if you type in a space the cursor jumps to the left side of the text row.

garvankeeley: ok thanks!

Issue #717, opened 2017-02-13
Webview auto zooming to incorrect location on selecting text field


jhreis:
Description:

When selecting 'username' on https://www.furaffinity.net/login/, the page auto-zooms into the text field, and then immediately to the top left of the page.

(reported via [Twitter](https://twitter.com/SwiftOnSecurity/status/830942106147553282))

**Did you search for similar issues before submitting this one?**
Yes

**Expected behavior:**
Zoom to the correct location on the page.

- Device (iPhone5, iPhone6s plus, iPad 3, ?):
iPhone 6s +

- Brave Version:
1.3

- Steps to reproduce:
    1. Navigate to https://www.furaffinity.net/login/
    2. Select the top field (username)

** Website problems only:
- did you check with Brave Shields down?
- did you check in Puffin/Dolphin/iCab/Maxthon (UIWebView-based browsers)?

- Screenshot if needed:

- Any related issues:

PR #719, opened 2017-02-13
Fix: #717: No longer adjust insets while zooming.

jhreis:


garvankeeley:
r+. Code looks nice and simple to check for zooms

PR #720, opened 2017-02-14
Fix issue with privacy mode tabs styles

jamesmudgett:
Privacy mode tabs titles and icons are black on dark grey, changed to light colors for more contrast.


garvankeeley:
fixes #722

PR #745, opened 2017-03-07
BrowserViewController 'coredata' has a viewDidAppear bug (ONLY on that branch)

garvankeeley:
Need to check on startup once only whether to restore tabs or to create a new tab. 
Bug was introduced during core data porting.

### DIFF
--- a/brave/src/frontend/BraveBrowserViewController.swift
+++ b/brave/src/frontend/BraveBrowserViewController.swift
@@ -44,9 +44,8 @@ class BraveBrowserViewController : BrowserViewController {
         // TODO: Should never call setupConstraints multiple times, this can cause huge headaches. Constraints should mostly be static with adjustments made to those constraints.
         setupConstraints()
 
-        struct RunOnceAtStartup { static var hasRun = false }
-        if !RunOnceAtStartup.hasRun {
-            RunOnceAtStartup.hasRun = true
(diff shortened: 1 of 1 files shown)


brave/src/frontend/BraveBrowserViewController.swift
```diff
@@ -43,14 +43,19 @@ class BraveBrowserViewController : BrowserViewController {
         
         // TODO: Should never call setupConstraints multiple times, this can cause huge headaches. Constraints should mostly be static with adjustments made to those constraints.
         setupConstraints()
-        if BraveApp.shouldRestoreTabs() {
-            tabManager.restoreTabs()
 
-            if tabManager.tabCount < 1 {
+        struct RunOnceAtStartup { static var hasRun = false }
```
jhreis: Thoughts about using GCD single dispatch over static var?
```
static dispatch_once_t once;
dispatch_once(&once, ^{
    if BraveApp.shouldRestoreTabs() { ...
        // ...
    }
});
```

garvankeeley:
It is main-thread so I tend to think that syntax suggests otherwise when reading code, but yeah, is cleaner to do that way, will switch, thanks!

garvankeeley:
Updated the PR

jhreis:
Really great. Truly appreciate it @garvankeeley!

PR #813, opened 2017-04-14
fix issue #808 - only show find in page on websites, hide on top sites

jamesmudgett:


Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -1003,6 +1003,19 @@ extension BrowserViewController: SearchViewControllerDelegate {
         updateFindInPageVisibility(visible: true)
         findInPageBar?.text = query
     }
+    
+    func searchViewControllerAllowFindInPage() -> Bool {
+        // Hides find in page for new tabs.
+        
+        if let st = tabManager.selectedTab {
```
jhreis: You can chain optional unwrappings, and simplify the return here:
```
if let st = tabManager.selectedTab, let wv = st.webView {
    return !AboutUtils.isAboutHomeURL(wv.URL)
}
return false
```

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -509,9 +510,14 @@ extension SearchViewController {
         case .SearchSuggestions:
             return 0
         case .BookmarksAndHistory:
-            return 0
+            return 1
         case .FindInPage:
-            return 22
+            if let sd = searchDelegate {
```
jhreis: Can use `where`, to remove the ternary and duplicate `0` return:
```
if let sd = searchDelegate where sd.searchViewControllerAllowFindInPage() {
    return 22
}
return 0
```

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -565,7 +575,12 @@ extension SearchViewController {
         case .BookmarksAndHistory:
             return data.count
         case .FindInPage:
-            return 1
+            if let sd = searchDelegate {
```
jhreis: Same as above, using `where`

Client/Frontend/Browser/SearchViewController.swift
jamesmudgett: Great! Will update

PR #868, opened 2017-05-10
Fix #858 - Added additional search engines

### LINKED ISSUE #858: Update Search Engines with Laptop
Description:
iOS currently is missing the following desktop supported search engines:
- searx.me
- Ecosia
- Mozilla Developer Network (MDN)
- Yandex

Yes

made2k:
PR for #858 

There were a few things done to help facilitate this PR. Yandex had already been added for certain locales (Russian, Turkish...) so simply just adding it to the `more` directory would cause duplicates when the device uses those languages settings. To counter this, I decided to use the file name as an identifier for each service (Using a property from the xml file would not be viable as things like shortName and URL can differ from region to region). 

Using the file name as an identifier allows the `moreEngineNames` to filter out any that have already been included. So future additions can have a general search engine in the more, but fine tune it for individual locales inside their associated directory.

A question with room for improvement is supplying the favicons. The problematic ones are Ecosia and MDN (retrieved from the image urls from [browser-laptop](https://github.com/cwmke/browser-laptop/blob/417a6bcc846ebb10ce1d2f1d5368f091488632a6/js/data/searchProviders.js)). They are just too small and show up pixelated. There's sites that provide alternative icons like icons8.com but I'm unsure of Braves attribution policy. I pulled the Yandex icon from the one used if it appears in the top sites section of a new tab (https://yastatic.net/iconostasis/_/k6R30TMf8fzC00yWQYc0EaJYc70.png)

I also truncated MDN from "Mozilla Developer Network (MDN)" to "Mozilla Developer" as it was truncated on every device in the search selection. That may be something for UX to decide if it's an acceptable trade off.


jhreis:
Thanks for the PR! 

Mostly just wanted to give you an update about this. We have been holding off on all feature tickets for the next release (1.4), as we have some significant changes coming, and want to minimize impact scope. Only merging scoped bug fixes and sync related items.

Thanks again 😄

made2k:
Thanks for the update. Waiting is definitely good, especially with some of the changes like the migration to Swift 3 (which will be very nice updating to the latest Xcode). Once those get in, I'll be able to update this PR with the changes to support Swift 3. After that, and getting sync released for iOS, we can worry about some features like this one.

Jacalz:
It is also worth noting that FindX was added to browser-laptop too 😅

made2k:
I'll look at updating this PR to be able to merge with all the changes that have gone in in the last couple months sometime this week or next depending on my schedule. I don't think the changes would be too involved which is good.

jamesmudgett:
Closing, please update and resubmit.

Issue #589, opened 2016-11-22
Fingerprint lock entire browser


garvankeeley:
Consider option to require fingerprint swipe to access the browser.

I assume this would lock on backgrounding (or have a timeout specified?) or screen lock

jamesmudgett:
In https://github.com/brave/browser-ios/tree/1.5.x

jamesmudgett:
Related https://github.com/brave/browser-ios/issues/1083

PR #881, opened 2017-05-27
Feature/browser lock

jamesmudgett:


jhreis:
Looks good at a glance. Currently not adding any new features for 1.4 (sync support), so this will be scheduled in 1.4.1, if sync goes out stable. 

Also, can you please create a ticket for this PR (pin code / fingerprint), so we can track release items (especially useful for QA).

jamesmudgett:
Ticket: https://github.com/brave/browser-ios/issues/589

brave/src/frontend/PinController.swift
```diff
@@ -0,0 +1,514 @@
...
+        
+        let pinViewSize = pinView.frame.size
+        let xEdge = (view.frame.width - pinViewSize.width) / 2
+        let yEdge = (view.frame.height - pinViewSize.height) / 2
+        
+        pinView.snp_makeConstraints { (make) in
+            make.edges.equalTo(CGPoint(x: xEdge, y: yEdge))
+        }
+        
+        title = "Set Pin"
```
jhreis: Localize

brave/src/frontend/PinController.swift
```diff
@@ -0,0 +1,514 @@
...
+        
+        var authError: NSError? = nil
+        let authenticationContext = LAContext()
+        if authenticationContext.canEvaluatePolicy(.DeviceOwnerAuthenticationWithBiometrics, error: &authError) {
+            authenticationContext.evaluatePolicy(
+                .DeviceOwnerAuthenticationWithBiometrics,
+                localizedReason: "Only awesome people are allowed",
+                reply: { [unowned self] (success, error) -> Void in
+                    if success {
+                        if self.successCallback != nil {
```
jhreis: `self.successCallback?()` should do it.

brave/src/frontend/PinController.swift
```diff
@@ -0,0 +1,514 @@
...
+        
+        var indicatorViewFrame = pinIndicatorView.frame
+        indicatorViewFrame.origin.x = (frameWidth - indicatorViewFrame.width) / 2
+        indicatorViewFrame.origin.y = messageLabelFrame.maxY + 18
+        pinIndicatorView.frame = indicatorViewFrame
+        
+        var x: CGFloat = 0
+        var y: CGFloat = indicatorViewFrame.maxY + spaceY
+        
+        for i in 0..<buttons.count {
```
jhreis: Probably not a big deal as this is already written, but it would probably be easier and more reliable to use stack views to create a table, instead of manually adjusting all of these frames.

We use similar layouts elsewhere (e.g. code words for sync), we could create a table class that is used to layout these types of views.

PR #897, opened 2017-06-06
Added support for True Key app extension

fsoleil:
Added support for True Key app extension to use it as password manager.

About True Key : https://www.truekey.com/


fsoleil:
Great, thanks a lot

PR #946, opened 2017-07-06
Swift3 - Pulling in Changes from Master since original Fork

jhreis:
Handling merges from master changes.


jhreis:
@jamesmudgett Wouldn't work against 1.4.x or master, since this includes _all_ of the Swift3 changes, so it would be ~1800 files, basically same as the Swift3 -> master PR. I have some changes for this PR, but agree, ultimately going to be hard to distinguish.

PR #1061, opened 2017-08-15
fix #1045 - hide password manager button always when in search mode

### LINKED ISSUE #1045: URL bar display issue
Yes

- Description:
URL bar, Password Picker, Cancel overlap.

iPad

- Brave Version:
1.4 (17.08.04.18)

- Steps to reproduce:
    1. Navigate to a page (like twitter.com) where you can login.
    2. Tap on 'Log in' button.
    3. Tap on the username field.
    4. Now tap on the URL bar.

- Actual Behavior
On the right hand side of the screen, the URL bar, password picker, and Cancel text overlap.

- Expected Behavior
I would expect that the password picker would have been hidden since the focus is now in the URL bar, then the display issue wouldn't occur.

Yes

Yes

Yes

- Extra QA steps:
    1.  Check both landscape and portrait modes for iPad.
    2. 
    3.

- Website problems only: 

- Screenshot if needed:
ipad_overlap

- Any related issues:

jamesmudgett:


brave/src/frontend/BraveURLBarView.swift
```diff
@@ -285,6 +285,8 @@ class BraveURLBarView : URLBarView {
         }
 
         if inSearchMode {
+            pwdMgrButton.isHidden = true
```
jhreis: Safe to assume `isHidden = false` is executed already somewhere else to recover from this?

jamesmudgett: Correct. Form tap uses a JS trick to test if form is a login form, this bubbles up to set to false elsewhere.

PR #1086, opened 2017-08-23
tab issues being fixed here: 

jamesmudgett:
1. tab titles on tab bar were not being restored on relaunch and creating a broken experience, these are now restored from CD. 
2.  tab insertion was removed to fix performance on relaunching app, with several tabs this was noticeable on older devices. Added a scheduling mechanism to throttle re-layout, tabs continue to be restored quickly while preserving the ux of tab insertion, expanded this to include tabs opened from javascript _blank targets which makes a better ux on sites like fb
3. zombie tab restores should no longer hit CoreData
4. preserve selected tab on restore, this was broken before.
5. keep sort order on tabs when launching new ones, requires a pass through open tabs and writes to CD (ignored on zombie tabs)
6. put back a feature that was removed regarding state of Browser and restoring, not using in commits here but original comment posted with commit was no longer relevant given the tabs title is being populated on launch. would be great to have this for other potential optimizations.

### DIFF
--- a/Client/Frontend/Browser/Browser.swift
+++ b/Client/Frontend/Browser/Browser.swift
@@ -253,9 +253,7 @@ class Browser: NSObject, BrowserWebViewDelegate {
         // we extract the information needed to restore the tabs and create a NSURLRequest with the custom session restore URL
         // to trigger the session restore via custom handlers
         if let sessionData = restorationData {
-            #if !BRAVE // no idea why restoring is needed, but it causes the displayed url not to update, which is bad
-                restoring = true
-            #endif
+            restoring = true
             lastTitle = sessionData.title
             if let title = lastTitle {
                 webView.title = title
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
@@ -828,7 +828,8 @@ class BrowserViewController: UIViewController {
             switchBrowsingMode(toPrivate: true)
         }
         
-        tabManager.addTabAndSelect(request)
+        //tabManager.addTabAndSelect(request)
+        tabManager.addAdjacentTabAndSelect(request)
     }
 
     func openBlankNewTabAndFocus(isPrivate: Bool = false) {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
@@ -828,7 +828,6 @@ class BrowserViewController: UIViewController {
             switchBrowsingMode(toPrivate: true)
         }
         
-        //tabManager.addTabAndSelect(request)
         tabManager.addAdjacentTabAndSelect(request)
     }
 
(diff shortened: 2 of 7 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -281,6 +291,7 @@ class TabManager : NSObject {
     
     fileprivate func restoreTabsInternal() {
         var tabToSelect: Browser?
+        isRestoring = true
```
jhreis: Rather than using this, we should just synchronize this class.

brave/src/frontend/tabsbar/TabWidget.swift
```diff
@@ -76,7 +76,9 @@ class TabWidget : UIView {
 
         close.addTarget(self, action: #selector(clicked), for: .touchUpInside)
         title.addTarget(self, action: #selector(selected), for: .touchUpInside)
-        title.setTitle("", for: .normal)
+        
+        let t = TabMO.getByID(browser.tabID ?? "", context: DataController.shared.mainThreadContext)
+        title.setTitle(t?.title, for: .normal)
```
jhreis: Super great

brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -200,6 +201,23 @@ class TabsBarViewController: UIViewController {
...
         t.remakeLayout(tabs.last?.spacerRight != nil ? tabs.last!.spacerRight : self.spacerLeftmost, width: w, scrollView: scrollView)
         tabs.append(t)
+        
+        if let index = at, index > -1 && index < tabs.count {
+            // Ignore all of this on bootup
+            if !getApp().tabManager.isRestoring && !insertTabScheduled {
+                // Trottle layout. Reduce re-layout bottleneck on fast tab creation.
+                insertTabScheduled = true
+                weak var weakSelf = self
+                postAsyncToMain(0.2) {
```
jhreis: Curious about the need for this.

Client/Frontend/Browser/Browser.swift
```diff
@@ -253,9 +253,7 @@ class Browser: NSObject, BrowserWebViewDelegate {
         // we extract the information needed to restore the tabs and create a NSURLRequest with the custom session restore URL
         // to trigger the session restore via custom handlers
         if let sessionData = restorationData {
-            #if !BRAVE // no idea why restoring is needed, but it causes the displayed url not to update, which is bad
-                restoring = true
-            #endif
+            restoring = true
```
jhreis: This was in a `!BRAVE` so should just be removed

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -828,7 +828,8 @@ class BrowserViewController: UIViewController {
             switchBrowsingMode(toPrivate: true)
         }
         
-        tabManager.addTabAndSelect(request)
+        //tabManager.addTabAndSelect(request)
```
jhreis: remove

jhreis:
Referencing https://github.com/brave/browser-ios/commit/a15f73cd7c66b343b2589fdb6ff16d7c8969487c as a problematic removal of code. Removed initially for restoration optimizations, but contained negative UX side effects.

PR #1121, opened 2017-08-24
fix issue where private message not displayed on ipad

jamesmudgett:


Client/Frontend/Home/TopSitesPanel.swift
```diff
@@ -174,6 +180,30 @@ class TopSitesPanel: UIViewController {
...
+            }
+            else {
+                make.left.equalTo(0)
+                make.right.equalTo(0)
+            }
+        }
+        
+        privateTabLinkButton.snp.makeConstraints { (make) -> Void in
+            make.top.equalTo(self.privateTabInfoLabel.snp.bottom).offset(10)
+            make.left.equalTo(0)
```
jhreis: Just an FYI, you can chain these together:
`make.left.right.bottom.equalTo(0)`

PR #1123, opened 2017-08-24
fix issue where icon can be covered on snackbar

jamesmudgett:


Client/Frontend/Widgets/SnackBar.swift
```diff
@@ -199,9 +199,9 @@ class SnackBar: UIView {
 
         textLabel.snp.remakeConstraints { make in
             make.top.equalTo(contentView)
-            make.width.equalTo(contentView)
+            make.left.equalTo(self.imageView.snp_right)
             make.trailing.equalTo(contentView)
-            make.bottom.lessThanOrEqualTo(contentView.snp.bottom)
+            make.bottom.lessThanOrEqualTo(contentView.snp_bottom)
```
jhreis: `_` api is deprecated, any reason the `.` was having issues? Should be functionally equivalent.
(same for above too)

PR #1130, opened 2017-08-25
remove vertical line next to new tab button on iPad, update background…

jamesmudgett:
remove vertical line next to new tab button on iPad, update background color to match the inactive tabs, cleanup weak code

### DIFF
--- a/brave/src/frontend/tabsbar/TabsBarViewController.swift
+++ b/brave/src/frontend/tabsbar/TabsBarViewController.swift
@@ -48,27 +48,17 @@ class TabsBarViewController: UIViewController {
 
         if UIDevice.current.userInterfaceIdiom == .pad {
             plusButton.setImage(UIImage(named: "add")!.withRenderingMode(.alwaysTemplate), for: .normal)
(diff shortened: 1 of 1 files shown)


brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -207,13 +197,12 @@ class TabsBarViewController: UIViewController {
...
                 // Trottle layout. Reduce re-layout bottleneck on fast tab creation.
                 insertTabScheduled = true
-                weak var weakSelf = self
-                postAsyncToMain(0.2) {
-                    weakSelf?.insertTabScheduled = false
-                    weakSelf?.isAddTabAnimationRunning = false
-                    weakSelf?.moveTab(t, index: index)
-                    weakSelf?.recalculateTabView()
-                    weakSelf?.updateSeparatorLineBetweenTabs()
+                postAsyncToMain(0.2) { [weak self] in
```
jhreis: This is nifty, haven't seen this before.

jhreis: Also, this does not need to be weak, as the block itself is not retained.

Issue #1135, opened 2017-08-28
Launch Delays


jhreis:
- Did you search for similar issues before submitting this one?
Yes

- Description:
Launching Brave application takes far too long

- Device (iPhone5, iPhone6s plus, iPad 3, ?):
All

- Brave Version:
1.4.1 (was introducing 1.4)

- Steps to reproduce:
    1. Make sure Brave is removed from memory (e.g. swipe up)
    2. Launch Brave
    3. Observe slow launch times

- Actual Behavior
Launches slow

- Expected Behavior
Launch fast

- Is this an issue with Beta build?
N/A

- Is this an issue in the currently released version?
Yes

- Can this issue be consistently reproduced?
Yes

- Extra QA steps:
    1. 
    2. 
    3.

- Website problems only: 
- did you check with Brave Shields down? 
- did you check in Puffin/Dolphin/iCab/Firefox (UIWebView-based browsers)? 

- Screenshot if needed:

- Any related issues:
https://github.com/brave/browser-ios/issues/1064: This was in response to _massive_ launch delays. It has greatly improved beyond this, but still needs much more work.

jamesmudgett:
Related: https://github.com/brave/browser-ios/issues/1192

PR #1140, opened 2017-08-30
Performance updates (FastImageCaching, restore tabs from launch)

jamesmudgett:
(then destroy once tab limit is hit), instead restored tabs will only create webviews on selectTab action, this creates a much smaller memory footprint and allows for opening large amounts of tabs to happen quickly. There is other room for improvement. Tabsbarviewcontroller needs to be refactored. Added logic to ignore layout code when setting is set to "never" resulting in almost instant loads of 88 tabs.

- Ignores web session restore on tabs as previous logic would create tabs and restore data then once limit was hit would remove the tab webview and kill tab.
- Fixes dead tabs #767 and #708
- Improves #1135, but doesn't eliminate all problems. Rewriting TabsBarViewController with UICollectionView should fix as currently the layout is still causing performance issues.
- Added FastImageCaching which can be used in conjunction with objectId on CD objects to store record related data, in our case Tabs. 
- Changed default background color on TopViewController to white.


Client/Frontend/Browser/TabManager.swift
```diff
@@ -289,6 +294,10 @@ class TabManager : NSObject {
         }
     }()
     
+    func forceRestoreTabs() {
```
jhreis: Concerned about having open access to `restoreTabsInternal`, as this getting called multiple times leads to terrible behavior. Any reason why we can't use force dispatch once on `restoreTabs`?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -96,6 +96,7 @@ class TabManager : NSObject {
     fileprivate let defaultNewTabRequest: URLRequest
     fileprivate let navDelegate: TabManagerNavDelegate
     fileprivate(set) var isRestoring = false
+    fileprivate var skipWebviewGeneration = false
```
jhreis: This seems highly error prone, storing a value on this level and using it for restoration. I think tab restoration should be very linear and have no ability to run multiple times, which would make it much more predictable.

I understand this is used outside of an isolated class, but I think we need to find a way to keep the call stack completely serial.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -303,6 +312,9 @@ class TabManager : NSObject {
         //  has been related to layout constraints on the tab tray (re-arranging tabs as they are being created)
         //  Since `move` recalculates each pre-existing tab's position. Hence the forced order here.
         let savedTabs = TabMO.getAll()
+        if savedTabs.count > BraveUX.MaxTabsInMemory {
```
jhreis: Could this ever be false? I think it's probably problematic if `restoreTabsInternal` can be called multiple times, and the first time this would always be true.

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -275,10 +275,17 @@ class DebugSettings: Setting, XMLParserDelegate {
...
         let string = try? String(contentsOf: url)
         let urls = JSON(parseJSON: string!)["links"].arrayValue.flatMap { URL(string: $0.stringValue) }
-        for url in urls {
-            let request = URLRequest(url: url)
-            let new = getApp().tabManager.addTab(request)
-            getApp().tabManager.selectTab(new)
+        let context = DataController.shared.workerContext
+        context.perform {
+            for url in urls {
+                let tabID = TabMO.freshTab()
```
jhreis: Since this is being dispatched in the background, it probably makes sense to give a context here to keep it all contained inside of a single thread.

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -275,10 +275,17 @@ class DebugSettings: Setting, XMLParserDelegate {
...
+                let data = SavedTab(id: tabID, title: url.baseDomain ?? url.absoluteString, url: url.absoluteString, isSelected: false, order: -1, screenshot: nil, history: [url.absoluteString], historyIndex: 0)
+                TabMO.add(data)
```
jhreis: Same here, probably should be passing in the context this is running in.

jamesmudgett: Great catch

Client/Frontend/Browser/TabManager.swift
jamesmudgett: Can be false if restoring less than (9) max tabs in which case we can create webviews for all tabs.

Client/Frontend/Browser/TabManager.swift
jamesmudgett: Restore tabs isn't necessarily a destructive process, it simply grabs the stored CD Tabs then calls addTab within tabManger and restores history, title, etc. 

You did bring up an issue which I may have missed. On calling this we should kill/remove existing tabs.

Client/Frontend/Browser/TabManager.swift
jamesmudgett: I think you're covering two topics with this comment. 

SkipWebviewGeneration is set when checking the number of stored tabs in CD, if it's above the max tabs allowed we don't create any webviews with exception of selectedTab.

There are only two ways I can see creating this limit. The other would be to pass a flag through the function but would be even more error prone as it could be passed by other classes that access tabManager. This at least only works within the restoreTabsInternal function and is set to false once iterating over stored tabs.

Client/Frontend/Settings/AppSettingsOptions.swift
jamesmudgett: I'm starting to catch a few problems with this concept, I'm going to strip or modify a bit more and update pr.

Client/Frontend/Browser/TabManager.swift
jamesmudgett: There are other issues related to restoring _all_ tabs when tabbar is already setup, would need to remove tabs then add, this is inefficient and causes a bunch of constraints to be updated which is slow. I'll go back to original method for now as this is somewhat unrelated to the purpose of the PR.

brave/src/util/ImageCache.swift
```diff
@@ -0,0 +1,135 @@
...
+                size = CGSize(width: size.height, height: size.width)
+            }
+            return size
+        }
+    }
+    
+    override init() {
+        super.init()
+        
+        let imageFormat = FICImageFormat()
```
jhreis: Just move all of this init setup into the `bitmapCache` variable.

brave/src/util/ImageCache.swift
```diff
@@ -0,0 +1,135 @@
...
+        let entity = ImageEntity(url: url)
+        let format = ImageFormatFrameDeviceFull
+        if bitmapCache.imageExists(for: entity, withFormatName: format) {
+            bitmapCache.deleteImage(for: entity, withFormatName: format)
+        }
+    }
+    
+    fileprivate func resize(_ image: UIImage, size: CGSize) -> UIImage {
+        let ratio = size.width / size.height
+        let height = size.height > image.size.height ? image.size.height : size.height
```
jhreis: `height = min(size.height, image.size.height)`

Client/Frontend/Browser/TabManager.swift
```diff
@@ -396,25 +418,24 @@ class TabManager : NSObject {
...
+    func configureTab(_ tab: Browser, request: URLRequest?, zombie: Bool = false, useDesktopUserAgent: Bool = false, index: Int = -1, createWebview: Bool = true) {
         debugNoteIfNotMainThread()
         if (!Thread.isMainThread) { // No logical reason this should be off-main, don't add a tab.
             return
         }
         objc_sync_enter(self); defer { objc_sync_exit(self) }
         
-        limitInMemoryTabs()
+        // We don't actually create webviews on restore.
+        if createWebview {
```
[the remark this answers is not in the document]
jamesmudgett: I think you're right, could do away with this if statement.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -303,29 +306,32 @@ class TabManager : NSObject {
         //  has been related to layout constraints on the tab tray (re-arranging tabs as they are being created)
         //  Since `move` recalculates each pre-existing tab's position. Hence the forced order here.
         let savedTabs = TabMO.getAll()
+        var createWebviews = true
```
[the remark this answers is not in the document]
jamesmudgett: This is not a bad idea, part of the reason is that with <10 tabs we are forcing a reload on all tabs. I don't think this is actually restoring from an offline state and we are blowing out 9 cached images which require a re-render. I think I will go ahead an treat < 10 tabs the same and only load the first.

brave/src/util/ImageCache.swift
```diff
@@ -56,21 +56,32 @@ class ImageCache: NSObject, FICImageCacheDelegate {
...
+        get {
+            // Always know portrait size for screen. 
+            // FIC entity design requires specifying size associated with entity with each format.
+            // To simplify image cache for our use (screeshots on webview) enforce portrait sizing.
+            // Display differences happen when capturing at landscape. An alternative design approach
+            // would be to create both portrait and landscape sizes on each cache() call and allow FIC
+            // to return image based on device orientation. This would require a larger cache size
+            // and we'd still add to complexity of accessing image cache. A better pattern for alternatvie
+            // image formats may be to subclass imageCache, or to create a new instance and pass in format.
+            var size = CGSize(width: UIScreen.main.bounds.width, height: UIScreen.main.bounds.height)
```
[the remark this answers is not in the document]
jamesmudgett: Yes, I did.

brave/src/frontend/BraveUX.swift
```diff
@@ -111,7 +111,7 @@ struct BraveUX {
     static let BackForwardButtonLeftOffset = CGFloat(10)
 
     static let ProgressBarColor = UIColor(white: 0.0, alpha: 0.08)
-    static let ProgressBarDarkColor = UIColor(white: 1.0, alpha: 0.08)
+    static let ProgressBarDarkColor = UIColor(white: 1.0, alpha: 0.2)
```
jamesmudgett: progress bar on private mode was almost unnoticeable

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -189,7 +189,7 @@ class BrowserViewController: UIViewController {
             footerBackground?.addSubview(toolbar!)
             footer.addSubview(footerBackground!)
             
-            footer.layer.shadowOffset = CGSize(width: 0, height: -1)
+            footer.layer.shadowOffset = CGSize(width: 0, height: -0.5)
```
jamesmudgett: This is a design tweak, footer line seemed a little too thick.

brave/src/frontend/tabsbar/TabsBarViewController.swift
```diff
@@ -228,14 +253,14 @@ class TabsBarViewController: UIViewController {
...
-        if getApp().tabManager.tabCount < 1 {
-            return
-        }
-    }
     
     func overflowIndicators() {
+        // super lame place to put this, need to find a better solution.
+        plusButton.tintColor = PrivateBrowsing.singleton.isOn ? UIColor.white : UIColor.black
+        collectionView.backgroundColor = PrivateBrowsing.singleton.isOn ? UIColor(white: 0.0, alpha: 0.2) : UIColor(white: 0.0, alpha: 0.075)
+
```
jamesmudgett: Thoughts on what to do with lines above?

jhreis: For two items, I'd just do:
```
let isPrivate = PrivateBrowsing.singleton.isOn
plusButton.tintColor = isPrivate ? UIColor.white : UIColor.black
collectionView.backgroundColor = UIColor(white: 0.0, alpha: isPrivate ? 0.2 : 0.075)
```

For anymore, I'd break them into an if-else block.

jamesmudgett: Cool, was curious about placement in code as it's not idea to check this on every scroll. My thinking is to setup an observer on PrivateBrowsing switch but doesn't currently support notifications.

jamesmudgett:
Code moved into new https://github.com/brave/browser-ios/tree/development branch.

Issue #1161, opened 2017-09-12
New private tab info labels don't handle longer texts well(Polish language)


iccub:
- Did you search for similar issues before submitting this one?
Yes
- Description:
New private tab info text doesn't handle longer texts well. It works fine in landscape mode as there is more space available.

- Device (iPhone5, iPhone6s plus, iPad 3, ?):
ip7

- Brave Version:
1.4.2(17.09.08.16)

- Steps to reproduce:
    1. Set phone language to Polish
    2. Open new private tab

- Actual Behavior
See attached screenshot for better explanation. Title text is out of bounds and call to action button text if not fully visible

- Expected Behavior
Should look nice.

- Is this an issue with Beta build?
It's the same on app store and beta builds

- Is this an issue in the currently released version?
Yes

- Can this issue be consistently reproduced?
Yes

- Screenshot if needed:
![fullsizeoutput_9f](https://user-images.githubusercontent.com/6950387/30326369-a49000a4-97c8-11e7-9ff8-1553f998867c.jpeg)

PR #1162, opened 2017-09-12
Fix 1161, improving private tab info for longer texts

iccub:
### What was wrong
New private tab didn't look good in Polish language, some texts were of out screen edges, see #1161 
 for screenshots. 

### How I fixed it
Added some margins and word wrapping.

I also decided to make title text a bit wider than the info and button, I think it looks better than having the same width for every label. But please let me know if you prefer other dimensions.

To help you decide I made screenshots with 3 different options: same title width, wider title, narrower title, please decide which is better(doubled the title length for testing purposes)
zrzut ekranu 2017-09-12 o 17 42 07

### What is still wrong
It still looks bad on smaller iPhone, tested on iphone SE simulator, view is not scrollable

zrzut ekranu 2017-09-12 o 17 46 46

The question is how would you like to solve this? Two things came to my mind: adding a scrollview, but this may be an overkill, or reduce Brave logo size, maybe even slightly adjusting shields counter size.

Whole view could be also migrated to StackViews to stop worrying about layout constraints.


iccub:
Yea I'll try to do this over the weekend

iccub:
@jamesmudgett hey, I wrongly told you that there's no ScrollView in this view. Looks like everything is placed in CollectionView(which is based on ScrollView) so instead hiding the Brave logo I simply added missing bottom constraints and view can be scrolled when content is too big.

Is it ok?
zrzut ekranu 2017-09-15 o 20 20 40

PR #1217, opened 2017-10-04
Set tab title to "New Tab" when navigating back to new tab page.

SlayterDev:
When navigating back to new tab page the title from the webview comes back as nil which was not triggering the `setTitle` method. A simple check that we are navigating to the new tab page in the path of the url fixes this.


SlayterDev:
Fixes issue #1164

Issue #1212, opened 2017-10-03
Closing tab doesn't release memory


iccub:
- Did you search for similar issues before submitting this one?
Yes
- Description:
Memory doesn't seem to be released after closing tabs, which leads to huge memory consumption over time and app crashes.
- Device (iPhone5, iPhone6s plus, iPad 3, ?):
Tested on iP7, iPad Air 2
- Brave Version:
1.5, latest development version
- Steps to reproduce:
    1. Open XCode, run app and select 'debug navigator'(or press CMD + 7), select memory tab
    2. Open few tabs, heavy sites preferred, I tested on motorsports.com. 
    3. Close them all, one by one or all at once and notice that memory usage didn't drop by much.
    4. If you repeat 1-3 few times app will crash

This can be also done using one tab, clicking through some links, closing tab and opening new one, multiple tabs make it just faster to reproduce. After crash and tab restoration memory is not as high as before crash.

- Actual Behavior
Closing tab releases only small amount of memory.

- Expected Behavior
It should release more memory that it currently does.

- Is this an issue with Beta build?
Yes, it happens on production too.
- Is this an issue in the currently released version?
Can't debug memory on production but easily reproduced crash after load-close tabs few times.
- Can this issue be consistently reproduced?
Yes.

- Website problems only: 
- did you check with Brave Shields down? 
Yes, it's the same
- did you check in Puffin/Dolphin/iCab/Firefox (UIWebView-based browsers)? 
I tested the behavior on firefox iOS. On firefox this is also noticeable but the scale is really low. I mean, opening new tab in firefox takes few megabytes of memory, when on brave it's more like 50-150. 

For comparision, after 3 rounds of opening 15 motorsports tabs and closing them brave memory started at ~80MB and ended at ~1.2GB usage, on firefox started at ~80MB and ended at ~150MB, see screenshots below.

- Screenshot if needed:
Brave memory usage, iPhone 7
<img width="1025" alt="zrzut ekranu 2017-10-03 o 11 41 38" src="https://user-images.githubusercontent.com/6950387/31119197-71e62818-a830-11e7-842e-a0e3a7c97e17.png">

Firefox iOS memory usage, iPhone 7 simulator, couldn't build it on device, brave memory usage behaves similar on device/simulator
<img width="900" alt="zrzut ekranu 2017-10-03 o 11 45 34" src="https://user-images.githubusercontent.com/6950387/31119263-a20cd2a8-a830-11e7-821e-827202149958.png">

[...]

PR #1218, opened 2017-10-05
Improve memory usage when closing tabs

SlayterDev:
Make `TabBarCell`'s reference to the `Browser` object weak.
When the cell is removed from the collection view the browser was being retained by a strong reference cycle. This has improved releasing memory when closing tabs but there is still more memory being used after the tab is closed than before the tab was opened in the first place.


SlayterDev:
Fixes issue #1212 

Here is the improvement to memory this PR makes:

screen shot 2017-10-04 at 2 14 31 pm

PR #1219, opened 2017-10-05
Scroll animations

jamesmudgett:


jamesmudgett:
Using pop to make allow animation interruptions. Api is simple and more straight forward than CoreAnimation which could be helpful as we implement more gesture handling and animations into UI.

PR #1222, opened 2017-10-10
Revival of Unit/UI tests + making them work on iPads

iccub:
Repaired all tests, they work pretty good now, not 100% reliability as there is a lot of networking and the tests are not really separated from each other.

Tried to leave them as they were before.

Also added some conditionals to make tests work on iPads. Ran about 10 full test runs and they work most of the times on both devices. Tested on simulators, my phone has polish settings so it won't pass tests.

### Dev note
Changed all occurrences of `google.com` page loads to `google.ca` because .com tried to be smart and changed my location to poland despite iOS region settings.


jhreis:
Great to see some cleanup on the Unit Test front. Been on my mind for a long time. Really appreciate it!

PR #1226, opened 2017-10-11
Update ISSUE_TEMPLATE to the new look on browser-laptop and updates the README 

Jacalz:
Not much more to it 👍 

Closes:  #1227

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,35 +1,49 @@
-- Did you search for similar issues before submitting this one?
+<!--
+Have you searched for similar issues? We have received a lot of feedback and bug reports that we have closed as duplicates.
+Before submitting this issue, please visit our wiki for common ones: https://github.com/brave/browser-laptop/wiki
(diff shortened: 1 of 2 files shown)


srirambv:
Please log an issue and link the PR

Jacalz:
Happy to help 👍

jhreis:
I see this is already merged, but can @Jacalz please still address my question?

jhreis:
I also, think the beta question is still important.

srirambv:
Beta build question is still there. This is what the new layout looks like
image

jhreis:
Was actually just talking with @Jacalz about this. Is that question still relevant due to having the Brave Version? We can pretty easily identify if something is beta or not ourselves with that info. Thoughts @srirambv ?

srirambv:
No. We can exclude the beta version. But I'd suggest that we need to mention in template to provide full details including build number so that we don't have to ask when users report.

Jacalz:
I have opened a new PR that has some more information and such but I am a bit on the side of keeping the beta and release channels to match with browser-android-tab because it is needed there as far as I know 😅

Issue #1209, opened 2017-10-02
Add ui support for Iphone X


Jacalz:
- Did you search for similar issues before submitting this one?
Yes
- Description:
Brave should have support for Iphone X ui before the phone gets released. 
- Device (iPhone5, iPhone6s plus, iPad 3, ?):
Iphone X

- Screenshot if needed:
This is how it looks at the moment:
![namnlost](https://user-images.githubusercontent.com/25466657/31092287-2ec2bbc8-a7ae-11e7-8d29-c224da2e6f2b.jpeg)

![namnlost2](https://user-images.githubusercontent.com/25466657/31092291-31e359e8-a7ae-11e7-979e-710c350eb892.jpeg)

![namnlost3](https://user-images.githubusercontent.com/25466657/31092292-31f99d16-a7ae-11e7-9e3f-ac7730f93cec.jpeg)

jhreis:
We will not be supporting the iPhone X.

Jacalz:
This will be in 1.5 and it is fixed so closing

jhreis:
Closing, if there are layout issues file a new bug ticket.

kjozwiak:
Unfortunately the QA team doesn't own an iPhone X... I'm going to see if there's anyone under the "#testers" channel that owns an iPhone X and could verify this issue.

In the meantime, I'll test this within the simulator as well, but it would be nice to get this checked on a real iPhone X.

petemill:
Tested on real iPhone X in 17.11.06.14 and 17.11.07.17 - Brave looks perfect now. It has the most space dedicated to content out of all the browsers - so, thanks for that!

![img_0118](https://user-images.githubusercontent.com/741836/33144270-4e6eea60-cf71-11e7-95b4-7d2c86aac821.png)
![img_0120](https://user-images.githubusercontent.com/741836/33144271-4e879eb6-cf71-11e7-9d6b-24b22c4b7c82.png)
![img_0121](https://user-images.githubusercontent.com/741836/33144272-4ea17a2a-cf71-11e7-9ae5-96e0b6bb269f.png)

PR #1229, opened 2017-10-12
iPhone X UI fixes

SlayterDev:
Fixes issue #1209 

This pull request serves to fix many of the layout issues the iPhone X has introduced. Apple has provided views with a new property called `safeAreaInsets` which provides the proper insets to be used to deal with the top notch and the home indicator. It is important to note this property is only available on iOS 11+ so consideration must be taken when dealing with these insets for older versions. I have gone through the app and adjusted the views to fit within the "safe area" as per Apple's guidelines. The screenshots below will show the new adjustments.

screen shot 2017-10-12 at 9 54 42 am
screen shot 2017-10-12 at 9 54 54 am
screen shot 2017-10-12 at 9 55 00 am
screen shot 2017-10-12 at 9 55 08 am


iccub:
Hey, I looked at your PR, some feedback:

I feel like the app needs some bigger layout refactor to make it work on iPhoneX. Your fixes are good but it will require a lot of them to make it work well. Some screenshots with feedback after my check:

Buttons are too close to corners
zrzut ekranu 2017-10-12 o 18 55 42

Notch blocking tabs in horizontal mode
zrzut ekranu 2017-10-12 o 18 56 00

Search suggestions at the bottom, corners + there are some grey areas on 'search in page' view
zrzut ekranu 2017-10-12 o 18 56 39

Notch blocks top hits thumbnail
zrzut ekranu 2017-10-12 o 19 34 02

After scrolling page down, toolbar is not hidden completely
zrzut ekranu 2017-10-12 o 18 57 08

I fell like the back button on side panel could be closer to the edge as notch doesn't hide it
zrzut ekranu 2017-10-12 o 18 58 28

Label covered by notch, probably only scaling text issue, other controls look ok, in English lang at least
zrzut ekranu 2017-10-12 o 19 00 38

A tricky one. When you open up slider and change orientation, it renders badly.
zrzut ekranu 2017-10-12 o 19 34 08

The same when you go horizontal -> vertical, also 'New folder' could be closer to the edge.
zrzut ekranu 2017-10-12 o 19 34 17


Most of it looks like it can be fixed by some conditional iPhoneX checks, but I wonder if it is good approach.

I was thinking about alternative approach:
- Move url bar view to navigation controller, move toolbar to toolbar view, I think nav controller and toolbar provide iPhoneX support of out box. Currently toolbar and url bar are just views glued to the top and bottom of screen. 
- refactor bookmark/shield sliders. Navigation controller should be sacred and never overlapped. Currently both sliders move nav controller, making it hard to calculate for ipX. I would make sliders that go only for safe area(whole view minus nav+toolbar). Additionally I would prefer opening full VC for shields and bookmarks on iPhones in portrait. On Safari bookmarks slider opens full screen on iPhone portrait, it slides on iPhone horizontal and both iPad orientations. Tldr; make sliding animations only over safe area.

Just my thoughts, maybe @jamesmudgett @jhreis had some insights about iPX preparations already.

SlayterDev:
All good points. Looks like the bottom toolbar not hiding is intermittent since in most of my testing it hid just fine. I would like to test this against iOS 11.1 (which is what the iPhone X will ship with) as I believe it may solve some of these issues. In some of your examples where you suggest moving elements closer to the edge while the phone is in landscape I would be reluctant to try. For one thing it would be a bit hacky to do so and for another Apple advises against it. It appears (however ugly) Apple wants to indent content on the sides while in landscape.

See here: landscape margins

iccub:
Yea I was wrong about this closer to edge things, sorry. It's because settings button on bookmarks slider is already near the edge and it confused me visually. So looks like it's actually the settings button that needs to be pushed away from screen edge.

SlayterDev:
Looking in to it. There is a `safeAreaLayoutGuide` attached to `UIView`s that seems to handle it pretty gracefully.

jamesmudgett:
We should increase the side bar widths to account for the new safe area inset. Good points @iccub definitely agree with moving to a navigation controller instead of adding header and footer as browserViewController subviews, then getting this stuff free. There are other benefits to this like getting better header & footer hide/show handling with scrollview and overall a cleaner structure.

SlayterDev:
Haven't looked at the Brave panel yet but at least with the bookmarks panel it appears that the buttons at the top have their positions set based on the center X of the container view. This may make it difficult to conform these buttons to the safe area.

SlayterDev:
I adjusted the find in page bar to align to the safe area. The home indicator still sits on top of the bar in landscape mode but I feel like making the bar taller may make it look weird.

screen shot 2017-10-13 at 9 52 46 am

PR #1250, opened 2017-10-19
improve tab preserve history state when a page is still in a loading …

jamesmudgett:
…state and not registered in history

### DIFF
--- a/brave/src/data/TabMO.swift
+++ b/brave/src/data/TabMO.swift
@@ -145,13 +145,16 @@ class TabMO: NSManagedObject {
         
         var urls = [String]()
         var currentPage = 0
+        
+        tab.webView?.backForwardList.update()
+        
         if let currentItem = tab.webView?.backForwardList.currentItem {
             // Freshly created web views won't have any history entries at all.
             let backList = tab.webView?.backForwardList.backList ?? []
             let forwardList = tab.webView?.backForwardList.forwardList ?? []
-            let backListMap = backList.map { $0.URL.absoluteString }
+            var backListMap = backList.map { $0.URL.absoluteString }
             let forwardListMap = forwardList.map { $0.URL.absoluteString }
-            let currentItemString = currentItem.URL.absoluteString
+            var currentItemString = currentItem.URL.absoluteString
             
             debugPrint("backList: \(backListMap)")
             debugPrint("forwardList: \(forwardListMap)")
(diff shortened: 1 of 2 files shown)


brave/src/data/TabMO.swift
```diff
@@ -149,12 +149,42 @@ class TabMO: NSManagedObject {
...
+            
+            /* Completely ignore forward history when passing urlOverride. When a webpage
+               hasn't fully loaded we attempt to preserve the current state of the webview.
+               urls that are currently loading aren't visible in history or forward history.
+               Our work around here is to append the currently requested URL to the end of the
+               navigation stack, and ignore forward history (as it would be replaced on full load).
+               There should be a very narrow edgecase where user who navigates back has no active
+               cache for the back url and close the browser while page is still being loaded-
+               resulting in lost forward history. */
+
```
jhreis: I think it would be much cleaner to reverse this logic:

```
guard let urlOverride = urlOverride else {
          // Business as usual.
          urls = backListMap + [currentItemString] + forwardListMap
          currentPage = -forwardList.count
          return
}
```

Then we can just use `urlOverride` as unwrapped

brave/src/data/TabMO.swift
```diff
@@ -149,12 +149,42 @@ class TabMO: NSManagedObject {
...
+            /* Completely ignore forward history when passing urlOverride. When a webpage
+               hasn't fully loaded we attempt to preserve the current state of the webview.
+               urls that are currently loading aren't visible in history or forward history.
+               Our work around here is to append the currently requested URL to the end of the
+               navigation stack, and ignore forward history (as it would be replaced on full load).
+               There should be a very narrow edgecase where user who navigates back has no active
+               cache for the back url and close the browser while page is still being loaded-
+               resulting in lost forward history. */
+            
+            if let urlOverride = urlOverride, backListMap.contains(urlOverride) || forwardListMap.contains(urlOverride) {
```
jhreis: This condition is functionally equivalent to the last one, and the beginning of the middle condition here. So we should be able to just:

```
urls = backListMap + [currentItemString]

if urlOverride != currentItemString {
    urls.append([urlOverride])
}

```

to replace this entire if-else chain. (in combination with above comment).

jamesmudgett: I see, however then we'd need to nest this in, so may make code a little harder to read. Above comment won't necessarily work as we still want to execute the else for all other cases- both where urlOverride is set and others where it isn't.

brave/src/data/TabMO.swift
jamesmudgett: In the event the other conditions aren't met but still passing in urlOverride value still want to continue with default setup. So this probably won't be much help.

jamesmudgett: This method is being called when url has changed and when page has loaded. On change we pass in urlOverride as the new url not yet registered in the uiwebview but loaded into localhost, this forces the new url to be set into the SavedTab data along with history state. 

When the web view has loaded the page this method is called without the urlOverride since we can rely on the web view/history state.

PR #1289, opened 2017-11-03
fix #1286 - updated fingerprinting protection to include new svg, can…

### LINKED ISSUE #1286: fingerprinting protection bypass
this fix needs to be ported to iOS/Android: https://github.com/brave/browser-laptop/issues/11683. i think you should be able to just pull app/extensions/brave/content/scripts/blockCanvasFingerprinting.js from browser-laptop master.

jamesmudgett:
…vas, and webrtc types.

Original: https://github.com/brave/browser-laptop/blob/master/app/extensions/brave/content/scripts/blockCanvasFingerprinting.js

### DIFF
--- a/brave/src/page-hooks/FingerprintingProtection.js
+++ b/brave/src/page-hooks/FingerprintingProtection.js
@@ -26,7 +26,7 @@ function trapInstanceMethod (item) {
 }
 
 var methods = []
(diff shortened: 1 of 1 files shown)


jhreis:
https://github.com/brave/browser-ios/issues/1286 has since been closed, and is not to be implemented.

PR #1338, opened 2017-11-21
Fix dau first launch ever param.

iccub:
Check for "dau_stat" was done wrong.
First app launch is when there is no value in preferences.


jhreis:
Great catch!

PR #1346, opened 2017-11-23
Fix #1337, adding leading 0 to dau appVersion param.

### LINKED ISSUE #1337: Add leading `.0` to version numbers used for server pings
For server pings, verify all iOS version numbers follow exact format: x.x.x (so for 1.6, it should be passed up as 1.6.0)

Should be implemented: `BraveApp:updateDauStat` (or at least used here)

See web implementation:
https://github.com/brave/vault-analytics/commit/07657c3dce53ca9549c4eb8e3e286413d5f094ca

iccub:
Using the same regex as in vault-analytics.

### DIFF
--- a/analytics/DAU.swift
+++ b/analytics/DAU.swift
@@ -6,6 +6,7 @@ import Shared
 // TODO: Separate logger for this kind of work?
 private let log = Logger.browserLogger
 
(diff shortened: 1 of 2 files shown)


jhreis:
Nice, good job minimizing the impact of this, to just DAU pings. Clean.

PR #1399, opened 2018-01-11
Swift4 migration

iccub:
Everything seems to work, tested on simulators and iPhone 7, iPad Air 2, with iOS 11.

If there are some bugs it's usually due to lack of `@objc` in methods. Unfortunately it's not caught on compile time so we need to check manually each app's feature to see if it works ok.

A lot of deleted lines is caused by new SQLite version.


jamesmudgett:
Can you resolve any conflicts and rebase to 1.6?

iccub:
There are some hard to track issues and @jhreis  decided to not to go with the migration, I think we can close it for now.
We can still consider moving to swift4 for 2.0 release

jhreis:
Yeah, no point rn to move this codebase into 4.0, we can always revisit in the future.

PR #1402, opened 2018-01-15
Fix #1397, favorites folder functionality.

### LINKED ISSUE #1397: Allow user to customize top sites
Currently top sites use a frequency algorithm, and are not pinnable or changable.

We want to allow the user to completely customize this. The bookmark panel will now have a static, undeletable folder called "Top Sites" (or some such name). Bookmarks stored here will now be the _only_ source for new tab top sites. If this folder is empty, the use will have no top sites.

Top Sites should be converted to a collection view to allow for an undetermined amount of bookmarks, and should prioritize bookmark name > bookmark url. Top Sites no longer need to be deletable from the top sites page either.

- [ ] Add static, undeletable Top Sites bookmark folder
- [ ] Prefill Top Sites folder with default websites
- [ ] Update Top Sites panel to be a collection view (scrollable)
- [ ] Update Top Sites to pull solely from the Top Sites bookmark folder
- [ ] Present bookmark `name` with a fallback to bookmark `url` (currently only `url` is ever shown)

iccub:
This PR removes top sites functionality and adds customizable 'Favorites' folder instead.

Changes overview:
1. Renamed 'top sites' to 'favorites' where possible. Exception is `TopSitesPanel`, this should be better named as this panel also shows blocker counter and private browsing screen so it not 'favorites' only.
2. Added `isFavoritesFolder` column(boolean, default value=false) to `Bookmark` model.
3. Changed UICollection view layout to flow layout. This is much simpler and good enough for our use case. Could also put shields stats as a header of flow layout, skipped that step for now.
4. Favorites folder can't be deleted or moved, folders can't be added inside of favorites folder.
5. Adding preset favorites bookmarks works for new and existing users.
6. Removed most of old top sites code to make `TopSitesPanel` class cleaner, LOC went from 1000 to 300, with separate files for data source and decorator for ui collection view cells.

### Known issues
Collectionview doesn't scroll on iPads. This is some legacy issue, I can also reproduce it on development branch without my changes applied. This bug only applies to iPads when there is over 18 favorites added.

### todo/nice to haves

- [ ] Better error handling/logging
- [ ] Drag to reorder tiles, currently if you want to change its order you need to do it from bookmarks panel
- [ ] A simpler way to add a website to favorites, maybe through a share extension?
- [ ] Add region-dependent preloaded websites, currently all users get the same favorites.
- [ ] Inform existing users on first launch about new changes.


jamesmudgett:
Stylistically looks like topsites collection cells have changed in alignment and size. Can this be made to look the same as it currently does on production?

iccub:
Will improve that.

Previously, cell size was dynamically calculated based on view's frame, it was causing me some problems with dynamic number of rows(prev it was just hardcoded to be always 2 rows) and it felt over engineered so I threw it away and replaced with `UICollectionFlowLayout`.
It should be possible to tweak it to make it closer to previous look.

PR #1427, opened 2018-01-30
User referral program

iccub:


brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
...
+        let todayInMilliseconds = Date().timeIntervalSince1970
+
+        guard todayInMilliseconds > referral.checkDate else {
+            log.debug("Not enough time has passed for referral ping.")
+            return
+        }
+
+        service.checkIfAuthorizedForGrant(referral: referral) { initialized, error in
+            // We also remove referral data when no download id is found on server.
+            if (error != nil && error == .downloadIdNotFound) || initialized == true {
```
jhreis: `nil` check should not be necessary as we are performing an equality check.

jhreis: `initialized` is a Bool, so no need for `== true`

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
...
+class UserReferralProgram {
+    let hostPlistKey = "URP_HOST"
+    let apiKeyPlistKey = "API_KEY"
+
+    let service: UrpService
+    let prefs: Prefs
+
+    init?() {
+        func getPlistString(for key: String) -> String? {
+            return Bundle.main.infoDictionary?[key] as? String ?? nil
```
jhreis: `?? nil` is inferred, can remove it.

brave/src/analytics/user-referral-program/ReferralData.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        self.downloadId = downloadId
+        self.code = code
+        checkDate = Date().timeIntervalSince1970 + _30days
+    }
+
+    // MARK: - NSCoding
+    func encode(with aCoder: NSCoder) {
+        aCoder.encode(downloadId, forKey: "downloadId")
```
jhreis: Should probably move all of these keys (e.g. `downloadId`, `code`), to `fileprivate static` strings.

brave/src/analytics/user-referral-program/UrpService.swift
```diff
@@ -0,0 +1,89 @@
...
+
+enum UrpError {
+    case networkError, downloadIdNotFound, ipNotFound, endpointError
+}
+
+/// Api endpoints for user referral program.
+struct UrpService {
+    let host: String
+    let apiKey: String
+    let apiKeyParam = "api_key"
```
jhreis: Make `static` anything that doesn't change.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Shared
+import SwiftKeychainWrapper
+
+private let log = Logger.browserLogger
+
+class UserReferralProgram {
+    let hostPlistKey = "URP_HOST"
```
jhreis: `static` and `private` if possible

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
...
+    let apiKeyPlistKey = "API_KEY"
+
+    let service: UrpService
```
jhreis: `private` or `fileprivate` I think works for these.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
...
+        self.prefs = prefs
+
+        guard let urpService = UrpService(host: host, apiKey: apiKey) else { return nil }
+        self.service = urpService
+    }
+
+    func referralLookup() {
+        service.referralCodeLookup { refferal, _ in
+            guard let ref = refferal else {
+                log.info("No referral code found")
```
jhreis: Great general use of logger, and good log level choices too!

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
...
+            let archivedReferral = NSKeyedArchiver.archivedData(withRootObject: ref)
+
+            self.prefs.setObject(archivedReferral, forKey: ReferralData.prefsKey)
+        }
+    }
+
+    func pingIfEnoughTimePassed() {
+        guard let data = prefs.objectForKey(ReferralData.prefsKey) as Data?,
+              let referral = NSKeyedUnarchiver.unarchiveObject(with: data) as? ReferralData else {
+                log.error("Could not retrieve referral model from keychain")
```
jhreis: Remove `keychain` from logs.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Shared
+import SwiftKeychainWrapper
```
jhreis: This import, no longer needed, right?

brave/src/analytics/user-referral-program/ReferralData.swift
```diff
@@ -0,0 +1,51 @@
...
+import SwiftyJSON
+
+private let log = Logger.browserLogger
+
+class ReferralData: NSObject, NSCoding {
+
+    static let prefsKey = "prefsReferral"
+    private let _30days = Double(30 * 24 * 60 * 60)
+
+    let downloadId: String
```
jhreis: Check permission level (e.g. private/fileprivate)

brave/src/analytics/user-referral-program/UrpService.swift
```diff
@@ -0,0 +1,89 @@
...
+    let downLoadIdKeyParam = "download_id"
+    let sessionManager: SessionManager
+
+    init?(host: String, apiKey: String) {
+        self.host = host
+        self.apiKey = apiKey
+
+        guard let hostUrl = try? host.asURL(), let normalizedHost = hostUrl.normalizedHost else { return nil }
+
+        // Certificate pinning
```
jhreis: Great job!

brave/src/analytics/user-referral-program/UrpService.swift
```diff
@@ -0,0 +1,89 @@
...
+    func referralCodeLookup(completion: @escaping (ReferralData?, UrpError?) -> ()) {
+        guard let endPoint = try? "\(host)/promo/initialize/ua".asURL() else {
+            completion(nil, .endpointError)
+            return
+        }
+
+        let params = [apiKeyParam : apiKey]
+
+        sessionManager.urpApiRequest(endPoint: endPoint, params: params) { response in
+            log.debug("response: \(response)")
```
jhreis: Maybe make log a bit more verbose? e.g. "URP response: "

Client/Application/AppDelegate.swift
```diff
@@ -153,6 +153,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
             }
         })
 
+        // MARK: User referral program
+        if let urp = UserReferralProgram() {
+            let isFirstLaunch = self.getProfile(application).prefs.arrayForKey(DAU.preferencesKey) == nil
```
jhreis: Probably fine for now, but we definitely will need to abstract this at some point. Not new though.

brave/src/analytics/user-referral-program/UrpService.swift
iccub: XCLogger provides function names so I decided it will be enough to make log look like `referralCodeLookup > response:..`.
But we don't utilize the log library too much. so more verbose log strings should be better I agree.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
iccub: `initialized` is optional Bool(true or false if request is good, or nil when there's 404 or other network errors)

Client/Application/AppDelegate.swift
iccub: I considered adding separate prefs flag for urp check, but went with the same logic for checking first launch as in `DAU.swift`.
If we want to go with separate flag I could move everything from here to `UserReferralProgram` class, now I just didn't want to couple DAU prefs within Urp classes.

jhreis: Don't think we should use a diff flag. I just meant isolating the "first time launch" stuff. This is fine though.

diracdeltas:
assigned @flamsmark for security/privacy review

iccub:
Relevant documentation to help with the review can be found [in Alamofire security section](https://github.com/Alamofire/Alamofire/blob/master/Documentation/AdvancedUsage.md#servertrustpolicy).

brave/BraveInfo.plist
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>URP_HOST</key>
+	<string>https://laptop-updates-staging.herokuapp.com</string>
```
diracdeltas: this will be updated to the *.brave.com host for production, right?

Client/Application/AppDelegate.swift
```diff
@@ -153,6 +153,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
...
+            if isFirstLaunch {
+                urp.referralLookup()
+            } else {
+                urp.pingIfEnoughTimePassed()
```
diracdeltas: I think there's a bug here. The specification says that clients only ping `/promo/activity` once after 30 days has elapsed. If I'm not mistaken, this will ping every time the browser starts up after 30 days has elapsed.

brave/BraveInfo.plist
iccub: Yes, host, api_key and host are going to be changed with build scripts I think.

Client/Application/AppDelegate.swift
iccub: Inside of this method there's date check(temporarily disable so we can test it better in beta version), that's why I added `ifEnoughTimePassed` for more clarity.

diracdeltas: @iccub yup i saw that: `todayInSeconds > referral.checkDate`. However, this doesn't ensure that `pingIfEnoughTimePassed` is only called once. It appears to be called on startup for all dates later than referral.checkDate.

iccub: @diracdeltas  You are right, I didn't see in the urp document that we need to call it only once.
So I should not care about return value(whether `isFinalized` is true or false) at all. I'm not sure when `isFinalized` is set to false other than server date comparision.

Are there any cases where we should consider retrying the api call? I think network errors should be ok for retry.

iccub:
Added once-per-day-for-5 days api call retry, as we've talked yesterday.

brave/setup.sh
```diff
@@ -28,6 +28,13 @@ if [[ $app_id == com.brave.ios.browser* ]]; then
     echo "DEVELOPMENT_TEAM=$dev_team_id" >> xcconfig/local-def.xcconfig
     echo adding fabric
     echo "./Fabric.framework/run $(head -1 ~/.brave-fabric-keys) $(tail -1 ~/.brave-fabric-keys)" > build-system/.fabric-key-setup.sh
+
+    # using comma delimiter to escape forward slashes properly
+    sed -e s/FABRIC_KEY_REMOVED/$(head -1 ~/.brave-fabric-keys)/  BraveInfo.plist.template |
```
jhreis: Curious about the additions for other secret keys?

brave/src/analytics/user-referral-program/UrpLogsViewController.swift
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+class UrpLogsViewController: UIViewController {
+    @IBOutlet weak var logsTextView: UITextView!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        guard let logs = UserDefaults.standard.value(forKey: "urpLogs") as? String else {
```
jhreis: There should be a `string` user default method to just pull a string.

brave/src/analytics/user-referral-program/UrpService.swift
```diff
@@ -0,0 +1,99 @@
...
+            return
+        }
+
+        let params = [UrpService.apiKeyParam : apiKey]
+
+        sessionManager.urpApiRequest(endPoint: endPoint, params: params) { response in
+            log.debug("Referral code lookup response: \(response)")
+            let json = JSON(response.data as Any)
+
+             let referral = ReferralData(json: json)
```
jhreis: Remove extra space

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,125 @@
...
+
+        guard let host = getPlistString(for: UserReferralProgram.hostPlistKey),
+            let apiKey = getPlistString(for: UserReferralProgram.apiKeyPlistKey), let prefs = getApp().profile?.prefs else {
+            log.error("Urp init error, failed to get values from Brave.plist.")
+            return nil
+        }
+
+        self.prefs = prefs
+
+        guard let urpService = UrpService(host: host, apiKey: apiKey) else { return nil }
```
jhreis: If this is going to block the init, it should probably be before the `self.prefs` line above.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,125 @@
...
+
+            if initialized == true {
+                UrpLog.log("Got initialized = true from server, removing all data from device")
+            }
+
+            if counter == 0 {
+                UrpLog.log("Retry counter is down to 0, failed to get data from server, removing all data from device")
+            }
+
+            if error == .downloadIdNotFound || initialized == true || counter == 0 {
```
jhreis: Torn about this, we are technically duplicating this logic with the above logs. What do you think about setting up a bool at top of method:
`var removeUrpData = false`

and then if any of the `if` checks are `true` reset it `removeUrpData = true`

Then at the end this statement would just be `if removeUrpData { // remove }`

Pros/cons both ways, but if we add an additional check it only needs to be added in once place. Thoughts?

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,125 @@
...
+                self.prefs.removeObjectForKey(ReferralData.prefsKey)
+                self.prefs.removeObjectForKey(UserReferralProgram.urpDateCheckPrefsKey)
+                self.prefs.removeObjectForKey(UserReferralProgram.urpRetryCountdownPrefsKey)
+            } else {
+                UrpLog.log("Network error or isFinalized returned false, decrementing retry counter and trying again next time.")
+                // Decrement counter, next retry happens on next day
+                self.prefs.setInt(counter - 1, forKey: UserReferralProgram.urpRetryCountdownPrefsKey)
+                let _1dayInSeconds = Double(1 * 60 * 60)
```
jhreis: This is only 1 hour, needs to be multiplied by `24`

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -407,6 +407,20 @@ class BraveTermsOfUseSetting: Setting {
...
 // MARK: - DEBUG
 
+class UrpDebugSetting: Setting, XMLParserDelegate {
+
+    override var title: NSAttributedString? {
+        return NSAttributedString(string: "URP logs", attributes: [NSForegroundColorAttributeName: UIConstants.TableViewRowTextColor])
+    }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let sb = UIStoryboard(name: "UrpLogs", bundle: nil)
```
jhreis: Not a huge fan of using Storyboard here. Feel programmatic or maybe even a xib is fine. Given this is just for debugging and hugely unlikely to change much, probably okay.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
iccub: I like your alternative. I did a lot of shortcuts to not spend too much time on urp logs, will clean up all the logs code you mentioned in the review.

brave/setup.sh
iccub: My bad, had to combine all key replacement regexes and forgot to remove old line of code for fabric/mixpanel.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,129 @@
...
+
+        guard let checkDate = self.prefs.objectForKey(UserReferralProgram.urpDateCheckPrefsKey) as TimeInterval? else {
+            log.error("Could not retrieve check date from preferences.")
+            return
+        }
+
+        let todayInSeconds = Date().timeIntervalSince1970
+
+        // Commenting out date comparision for beta testing.
+        guard todayInSeconds > checkDate else {
```
jhreis: This should be a standard `if`, using `guard` to replace boolean algebra is a bit hard to read. 

`if todayInSections <= checkDate` is much much more clear IMO.

brave/src/analytics/user-referral-program/UrpLogsViewController.swift
```diff
@@ -0,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+
+struct UrpLog {
+    static let prefsKey = "urpLogs"
+
+    static func log(_ text: String) {
+        var logs = UserDefaults.standard.value(forKey: prefsKey) as? String ?? ""
```
jhreis: Isn't there a `string` function off of UserDefaults? (relevant for other places too)

brave/src/analytics/DAU.swift
```diff
@@ -83,6 +83,27 @@ struct DAU {
             + firstLaunchParam(isFirstLaunch)
             // Must be after setting up the preferences
             + weekOfInstallationParam
+
+        // MARK: User referrals
+
+        // Removing referral code after 90 days
+        if let referralCodeDeleteDate = prefs.objectForKey(ReferralData.PrefKeys.referralCodeDeleteDate) as TimeInterval?,
```
jhreis: I think this logic should be inside of the ReferralData class, keep all of it isolated, makes for easier auditing for changes. The DAU should just check if there is a value in there and use that, not having any control logic.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,137 @@
...
+                UrpLog.log("No referral code found")
+                return
+            }
+
+            self.prefs.setString(ref.downloadId, forKey: ReferralData.PrefKeys.downloadId)
+            self.prefs.setString(ref.referralCode, forKey: ReferralData.PrefKeys.referralCode)
+
+            UrpLog.log("Found referral: downloadId: \(ref.downloadId), code: \(ref.referralCode)")
+            // In case of network errors or getting `isFinalized = false`, we retry the api call but not more than once
+            // per day and not for more than for 5 days. After this time, download record is deleted.
```
jhreis: This commented outdated now, right? Probably best to use generics in comments "more than for X days", since they are hard to keep updated with the code.

brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -0,0 +1,137 @@
...
+            self.initRetryPingConnection(numberOfTimes: 30)
+        }
+    }
+
+    private func initRetryPingConnection(numberOfTimes: Int32) {
+        let _10minutes: TimeInterval = 10 * 60
+        // We don't want to wait 30 days for beta build testing.
```
jhreis: Same, probably best to remove the literal 30 day number from the comment
.

brave/src/analytics/DAU.swift
jhreis: In fact a more elaborate approach could be to instantiate multiple `initRetryPingConnection` with some callback to be executed daily (or for beta every launch).

One would be for the 30 day retry logic and another instance would be for the 90 day ref code.

Making it generic would be really nice, so we could even handle other timed executions for other time limited tasks. Just a thought, would require some more heavy refactoring.

brave/src/analytics/user-referral-program/UrpService.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        sessionManager = SessionManager(
+            serverTrustPolicyManager: ServerTrustPolicyManager(
+                policies: serverTrustPolicies
+            )
+        )
+    }
+
+    func referralCodeLookup(completion: @escaping (ReferralData?, UrpError?) -> ()) {
+        guard let endPoint = try? "\(host)/promo/initialize/ua".asURL() else {
```
jhreis: Should not be manually creating a URL like this. If host has a trailing `/` this will double slash it. Use URL's `appendPathComponent`

brave/src/analytics/user-referral-program/UrpService.swift
```diff
@@ -0,0 +1,99 @@
...
+            log.debug("Referral code lookup response: \(response)")
+            let json = JSON(response.data as Any)
+
+            let referral = ReferralData(json: json)
+            completion(referral, nil)
+        }
+    }
+    
+    func checkIfAuthorizedForGrant(with downloadId: String, completion: @escaping (Bool?, UrpError?) -> ()) {
+        guard let endPoint = try? "\(host)/promo/activity".asURL() else {
```
jhreis: Same as above. Use URL string path combinations.

PR #1435, opened 2018-02-10
Update ad-block dependency to above 3.1

Jacalz:
See https://github.com/brave/browser-laptop/issues/13092

This will result in better ad blocking thanks to new ad-block engine and better regional ad blocking for Swedish users!


Jacalz:
Gives a whole world of problems :(

PR #1463, opened 2018-02-26
Remove BraveInfo.plist from tracking by git.

lapcat:
The setup.sh script copies BraveInfo.plist.template to BraveInfo.plist, so the latter should no longer be in version control. There is currently a whitespace-only difference between BraveInfo.plist.template and BraveInfo.plist which causes git to indicate a file modification after setup.sh is run.

BraveInfo.plist was added to .gitignore in 57d6f33c5c05c19ced12b512a0eeebe2a6f47e8c. However, the ignore file only applies to files that are not already tracked. BraveInfo.plist was re-added to the repository in 8fff52c5f170b756a60fceffff175711554210fb, which means the .gitignore entry had no effect. Removing BraveInfo.plist from the repository will allow the .gitignore to properly take effect.


lapcat:
The file was modified on the 1.6 branch https://github.com/brave/browser-ios/commit/8167daecf2a450d1464769a480814bba651015b9#diff-bf1779099d00da4669f21a24d13636cb so there may be a merge conflict. That was the whitespace-only difference I mentioned. The file should still be deleted from the repository, though, because it is simply overwritten by the setup script and is intended to be ignored by git in the .gitignore.

jhreis:
This PR was never merged in to development yet, so no conflicts yet.

I will merge this into development, and then merge into 1.6. As you mentioned, this will cause a conflict, but that's not a big deal.

PR #1466, opened 2018-02-26
Sync updates 1.6

jamesmudgett:
- New codewords input supports any number of words
- Pop to sync settings screen
- Strings updates


Client.xcodeproj/project.pbxproj
```diff
@@ -5543,6 +5544,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEVELOPMENT_TEAM = "";
```
jhreis: Remove.

brave/src/frontend/sync/SyncAddDeviceTypeViewController.swift
```diff
@@ -96,7 +96,7 @@ class SyncAddDeviceTypeViewController: SyncViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        title = Strings.Sync
+        title = "Sync a Device"
```
jhreis: Should be in string file.

brave/src/frontend/sync/SyncAddDeviceViewController.swift
```diff
@@ -199,6 +231,19 @@ class SyncAddDeviceViewController: SyncViewController {
         if deviceType == .computer {
             SEL_showCodewords()
         }
+        
+        updateLabels()
+    }
+    
+    func updateLabels() {
+        if deviceType == .mobile {
+            titleLabel.text = modeControl.selectedSegmentIndex == 0 ? Strings.SyncAddDeviceScan : Strings.SyncAddDeviceWords
```
jhreis: Looks like `titleLabel` is same regardless of `deviceType`, right? If so, probably move out of `if` check.

brave/src/frontend/sync/SyncAddDeviceViewController.swift
```diff
@@ -207,16 +252,32 @@ class SyncAddDeviceViewController: SyncViewController {
...
     }
     
+    func SEL_copy() {
+        UIPasteboard.general.string = self.codewordsView.text
+        didCopy = true
+    }
+    
     func SEL_changeMode() {
-        barcodeView.isHidden = (modeControl.selectedSegmentIndex == 1)
+        qrCodeView.isHidden = (modeControl.selectedSegmentIndex == 1)
```
jhreis: This logic is starting to look a little weird, some are using `selectedSegmentIndex` and then the `copiedlabel` is based on a different hidden prop. Probably be worth creating a new bool:
`let isFirstIndex = modeControl.selectedSegmentIndex == 0`
And just using that throughout the next 4 `isHidden` setters.

brave/src/frontend/sync/SyncAddDeviceViewController.swift
```diff
@@ -207,16 +252,32 @@ class SyncAddDeviceViewController: SyncViewController {
...
+        
+        updateLabels()
     }
     
     func SEL_done() {
         // Re-activate pop gesture in case it was removed
         navigationController?.interactivePopGestureRecognizer?.isEnabled = true
         
-        self.navigationController?.popToRootViewController(animated: true)
+        if let viewController = self.navigationController?.viewControllers[1] {
```
jhreis: I don't think this check is safe, as it can still go out of bounds (the array index is not an optional).
Also, I fail to see how this would work in most cases, as `[1]` should always be true (as `0` is the root VC, so `1` would by default to `self` if there was a single VC on top (e.g. settings page). Maybe I'm missing something here, so may need clarification.

brave/src/frontend/sync/SyncCodewordsView.swift
```diff
@@ -2,132 +2,91 @@
...
     }
     
-    override func sizeToFit() {
-        let field = fields[fields.count - 1]
-        var f = frame
-        f.size.width = bounds.width
-        f.size.height = field.frame.maxY
-        frame = f
+    func wordCount() -> Int {
+        let words = field.text.components(separatedBy: " ")
```
jhreis: Rather than pulling directly from `field` again, just use `codeWords()`, this will also trim the whitespace to not accidentally add an additional word from the space separation.
If ^ words, then I think we can just use `return codeWords().count`, which may make this function unnecessary.

brave/src/frontend/sync/SyncCodewordsView.swift
```diff
@@ -2,132 +2,91 @@
...
-    func textFieldShouldReturn(_ textField: UITextField) -> Bool {
-        if textField.tag < fields.count - 1 {
-            let field = fields[textField.tag + 1]
-            field.becomeFirstResponder()
-        } else {
-            doneKeyCallback?()
-        }
-        return true
+    func textView(_ textView: UITextView, shouldChangeTextIn range: NSRange, replacementText text: String) -> Bool {
+        return text != "\n"
```
jhreis: Probably unimportant, but will not prevent pasting in content with new lines (probably not an important edge case though).

brave/src/frontend/sync/SyncPairCameraViewController.swift
```diff
@@ -22,12 +22,16 @@ class SyncPairCameraViewController: SyncViewController {
...
         super.viewDidLoad()
         
-        title = Strings.Pair
+        title = Strings.ScanSyncCode
         
         // Start observing, this will handle child vc popping too for successful sync (e.g. pair words)
         NotificationCenter.default.addObserver(forName: NSNotification.Name(rawValue: NotificationSyncReady), object: nil, queue: OperationQueue.main, using: {
             notification in
-            self.navigationController?.popToRootViewController(animated: true)
+            if let viewController = self.navigationController?.viewControllers[1] {
```
jhreis: Same as before.

brave/src/frontend/sync/SyncPairWordsViewController.swift
```diff
@@ -7,9 +7,26 @@ class SyncPairWordsViewController: SyncViewController {
...
     var scrollView: UIScrollView!
     var containerView: UIView!
-    var helpLabel: UILabel!
     var codewordsView: SyncCodewordsView!
     
+    lazy var wordCountLabel: UILabel = {
+        let label = UILabel()
+        label.font = UIFont.systemFont(ofSize: 13, weight: UIFontWeightRegular)
+        label.textColor = BraveUX.GreyE
+        label.text = "Word count: 0"
```
jhreis: Use string file version.

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -180,7 +180,7 @@ class SyncDeviceSetting: Setting {
     internal var device: Device
     
     internal var displayTitle: String {
-        return "\(device.deviceDisplayId ?? "") :: \(device.name ?? "")"
+        return "\(device.name ?? "")"
```
jhreis: Desktop currently shows this (or did anyway), fine removing it as long as it is intentional. If so, no longer need the wrapping `"\()"`

brave/src/frontend/sync/SyncCodewordsView.swift
```diff
@@ -2,132 +2,91 @@
...
-        let field = fields[fields.count - 1]
-        var f = frame
-        f.size.width = bounds.width
-        f.size.height = field.frame.maxY
-        frame = f
+    func wordCount() -> Int {
+        let words = codeWords()
+        var count = words.count
+        // Don't count if it's just a space (no characters entered for new word)
+        if count > 0, let last = words.last, last.count < 1 {
```
jhreis: Should not be needed anymore, as codeWords() removes all whitespace, so check should not be needed.

brave/src/frontend/sync/SyncAddDeviceViewController.swift
```diff
@@ -207,16 +254,35 @@ class SyncAddDeviceViewController: SyncViewController {
...
-        barcodeView.isHidden = (modeControl.selectedSegmentIndex == 1)
-        codewordsView.isHidden = (modeControl.selectedSegmentIndex == 0)
+        let isFirstIndex = modeControl.selectedSegmentIndex == 0
+        
+        qrCodeView.isHidden = !isFirstIndex
+        codewordsView.isHidden = isFirstIndex
+        copyPasteButton.isHidden = isFirstIndex
+        
+        if copyPasteButton.isHidden {
+            copiedlabel.isHidden = true
```
jhreis: This should also use `isFirstIndex`

brave/src/frontend/sync/SyncPairCameraViewController.swift
```diff
@@ -62,7 +63,12 @@ class SyncPairCameraViewController: SyncViewController {
                 }
                 
                 Scanner.Lock = true
-                self.cameraView.cameraOverlaySucess()
+                DispatchQueue.main.async {
```
jhreis: Not needed, `splitBytes` runs on current thread.

brave/src/frontend/sync/SyncPairWordsViewController.swift
```diff
@@ -95,13 +125,18 @@ class SyncPairWordsViewController: SyncViewController {
...
+        codewordsView.becomeFirstResponder()
     }
     
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
-        
+    func SEL_paste() {
+        if let contents = UIPasteboard.general.string {
+            // remove linebreaks and whitespace, split into codewords.
+            codewordsView.setCodewords(data: contents.trimmingCharacters(in: .whitespacesAndNewlines).components(separatedBy: " "))
```
jhreis: Should probably be made a string extension (stringextensions.swift), since we use in two places.

brave/src/frontend/sync/SyncAddDeviceViewController.swift
```diff
@@ -207,16 +254,35 @@ class SyncAddDeviceViewController: SyncViewController {
...
+        
+        updateLabels()
     }
     
     func SEL_done() {
         // Re-activate pop gesture in case it was removed
         navigationController?.interactivePopGestureRecognizer?.isEnabled = true
         
-        self.navigationController?.popToRootViewController(animated: true)
+        if let viewController = self.navigationController?.viewControllers[1], viewController.isKind(of: SyncSettingsViewController.self) {
```
jhreis: This code is just weird and requires a good amount of tight coupling for the code to understand what is going on. I think a callback block pass through would be better. e.g. setup a SyncSettingsScreen protocol with property "nextStep` that is fired whenever the next step in the setup is required. We can then also just pass that through to each screen. Notifications and weird viewController index pops are just really fragile and require a decent amount of investigation to identify issues.

Alternatively using delegate / protocol would be much nicer, just pass through a delegate to all settings screens, using notification center here where logic should be completely unidirectional is pretty bloated.

brave/src/frontend/sync/SyncPairCameraViewController.swift
```diff
@@ -22,13 +22,14 @@ class SyncPairCameraViewController: SyncViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        title = Strings.Pair
-        
         // Start observing, this will handle child vc popping too for successful sync (e.g. pair words)
         NotificationCenter.default.addObserver(forName: NSNotification.Name(rawValue: NotificationSyncReady), object: nil, queue: OperationQueue.main, using: {
             notification in
-            self.navigationController?.popToRootViewController(animated: true)
+            NotificationCenter.default.post(name: NotificationPushToSyncSettings, object: nil)
```
iccub: This handler sometimes skips showing qr code/codewords VC and proceeds to settings screen immediately. I don't see why this observer is needed at all, probably some legacy stuff?

brave/src/frontend/sync/SyncCodewordsView.swift
```diff
@@ -2,132 +2,91 @@
...
-            fieldFrame.origin.y = y
-            fieldFrame.size.width = w
-            fieldFrame.size.height = h
-            field.frame = fieldFrame
-            
-            x = x + w
-        }
+    func codeWords() -> [String] {
+        let text = field.text.trimmingCharacters(in: .whitespacesAndNewlines)
+        return text.components(separatedBy: " ")
```
iccub: Doesn't count properly with multiple spaces strings like `abc_______def`

brave/src/frontend/BraveSettingsView.swift
```diff
@@ -181,6 +190,16 @@ class BraveSettingsView : AppSettingsTableViewController {
     
         return settings
     }
+    
+    func pushSyncSettings() {
```
iccub: This can trigger multiple times. To reproduce, open 'Scan a sync code', toggle few times between camera vc and code words vc, on enter sync code words tap 'back' few times. Then scan a qr code, and you will see that sync settings screen is being popped few times.

brave/src/frontend/sync/SyncPairWordsViewController.swift
```diff
@@ -74,13 +94,23 @@ class SyncPairWordsViewController: SyncViewController {
...
+            make.top.equalTo(codewordsView.snp.bottom)
+            make.left.equalTo(codewordsView).inset(24)
+        }
+        
+        copyPasteButton.snp.makeConstraints { (make) in
+            make.top.equalTo(codewordsView.snp.bottom)
+            make.right.equalTo(codewordsView).inset(24)
+        }
+        
+        useCameraButton.snp.makeConstraints { (make) in
```
iccub: This button is not visible on smaller iPhones, need to adjust views a bit.

brave/src/frontend/sync/SyncAddDeviceTypeViewController.swift
jamesmudgett: Was addressed in https://github.com/brave/browser-ios/pull/1466/commits/51d3f8c4312f4b04ac42a213d85d66de489551d2

jhreis: In what way? Reason we are leaving it?
See: https://github.com/brave/browser-ios/pull/1466/files#diff-1ffcfecca234a551e0614f4b5ab1f2efR21

brave/src/frontend/sync/SyncCodewordsView.swift
jamesmudgett: New lines should be removed from clipboard data on paste from SyncPairWordsViewController

PR #1468, opened 2018-02-27
Better navigation for sync screens.

iccub:


brave/src/frontend/sync/SyncSettingsViewController.swift
```diff
@@ -180,6 +184,11 @@ class AddDeviceSetting: Setting {
 
     override func onClick(_ navigationController: UINavigationController?) {
         let view = SyncAddDeviceTypeViewController()
+        view.popHandler = {
+            if let settingsVC = navigationController?.childViewControllers[1] {
+                navigationController?.popToViewController(settingsVC, animated: true)
```
jhreis: Since this _is_ the SyncSettingsVC you can just use `self` instead.

`navigationController?.popToViewController(self, animated: true)`

brave/src/frontend/sync/SyncWelcomeViewController.swift
```diff
@@ -120,10 +120,26 @@ class SyncWelcomeViewController: SyncViewController {
     }
     
     func newToSyncAction() {
-        navigationController?.pushViewController(SyncAddDeviceTypeViewController(), animated: true)
+        let addDeviceTypeViewController = SyncAddDeviceTypeViewController()
+        addDeviceTypeViewController.popHandler = {
+            self.pushSyncSettingsViewController()
```
jhreis: A bit confused about this, let's chat about it later, not sure why we are pushing a VC on a pop action.

brave/src/frontend/sync/SyncSettingsViewController.swift
iccub: This is actually `AddDeviceSetting` class not, `SyncSettingsViewController`. And first child controller of the navigation controller points to main settings menu, so we need to use that bad array[1] index which looks bad.

jhreis: This is wrong I think. This ends up popping back to the welcome screen.
e.g. welcome -> new sync -> add device -> add device type -> sync settings -> add device -> add device type -> pops back to welcome screen, which makes sense, that is [1] afterall.

Everything should just do a normal pop back until we reach the sync settings page, when that pops the add device onto it, it then needs to set a `popToViewController(self)`

iccub: This `AddDeviceSetting` button is placed only in SyncSettings vc so it always goes back to settings vc, for WelcomeScreen vc there are separate buttons, Welcome screen is regular view controller where SyncSettings a view controller is based on `AppSettingsTableViewController` , we definitely need to look at this code together 😄

brave/src/frontend/sync/SyncAddDeviceViewController.swift
```diff
@@ -274,13 +275,8 @@ class SyncAddDeviceViewController: SyncViewController {
...
     
     func SEL_done() {
-        // At this point we're not sure if we started from welcome screen or sync settings vc
-        // SyncSettings may not be on the stack. Alternatively, chaining back to BraveSettings through
-        // references adds significantly more complexity and some loss of context clarity.
-        // The drawback is that this check is needed in two places. Durring add and after joining existing chain.
-        
-        if let syncSettingsView = navigationController?.viewControllers.first(where: { $0.isKind(of: SyncSettingsViewController.self) }) {
-            navigationController?.popToViewController(syncSettingsView, animated: true)
+        if let handler = popHandler {
```
jhreis: This logic shouldn't be duplicated either, we may need to find a similar parent vc to handle this logic, the child does not always need to handle pushing the next level.

PR #1475, opened 2018-03-02
Fix #1465, update ad block region parser.

### LINKED ISSUE #1465: setup.sh Error:  Cannot find module './node_modules/ad-block/lib/regions.js'
When I run "brave/setup.sh" I get the following error:

```
module.js:540
    throw err;
    ^

Error: Cannot find module './node_modules/ad-block/lib/regions.js'
    at Function.Module._resolveFilename (module.js:538:15)
    at Function.Module._load (module.js:468:25)
    at Module.require (module.js:587:17)
    at require (internal/module.js:11:18)
    at [eval]:1:1
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at Object.runInThisContext (vm.js:139:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:643:30)
    at evalScript (bootstrap_node.js:462:27)
```

It appears that the "regions.js" file was deleted in https://github.com/brave/ad-block/commit/c66098d911eda3bd24065072d9a24404569af01e#diff-e8acc63b1e238f3255c900eed37254b8

iccub:
Ad block library changed region list from javascript json file to c++ array.
Build script needed an update.


brave/setup.sh
```diff
@@ -45,9 +45,7 @@ fi
 
 echo GENERATED_BUILD_ID=`date +"%y.%m.%d.%H"` >> xcconfig/build-id.xcconfig
 
-npm update
```
iccub: We don't need to update npm anymore, right?

jhreis: Yeah, we actually do, this is what updates the actual npm libs (e.g. adblock). If this can lead to a different adblock file things like that.

brave/setup.sh
```diff
@@ -47,7 +47,12 @@ echo GENERATED_BUILD_ID=`date +"%y.%m.%d.%H"` >> xcconfig/build-id.xcconfig
 
 npm update
 
-node -e "require('./node_modules/ad-block/lib/regions.js').forEach((x) =>{ if (x.lang) {console.log(x.lang + ',' + x.uuid)} } )" > adblock-regions.txt
+./build-system/get_adblock_regions
+
+if [ ! -e adblock-regions.txt ]; then
```
jhreis: I think this should probably check the actual input source that the script is using. This txt file will probably always exist, it just might be empty. lmk if my comment doesn't seem right.

Issue #1477, opened 2018-03-05
[hackerone] 321896


jumde:
https://hackerone.com/reports/321896

PR #1478, opened 2018-03-07
 Request confirmation from user before opening apps using custom urls

jumde:
See here: https://hackerone.com/reports/321896

fix https://github.com/brave/browser-ios/issues/1477


brave/src/webview/BraveWebView.swift
```diff
@@ -8,6 +8,9 @@ import JavaScriptCore
 let kNotificationPageUnload = "kNotificationPageUnload"
 let kNotificationAllWebViewsDeallocated = "kNotificationAllWebViewsDeallocated"
 
+// The list of URI schemes that can be opened in browser
+private let inBrowserURISchemes = ["about", "data", "file", "ftp", "http", "https"]
```
diracdeltas: do we want to support ftp?

diracdeltas: also , 'ws', 'wss:, 'blob'?

jumde: It works on the Desktop browser.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +172,18 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handleExternalURL (_ url: URL) -> Void {
+        let alertController = UIAlertController(title: "", message: "Allow '" + url.relativeString + "' to open external application?", preferredStyle: UIAlertControllerStyle.alert)
```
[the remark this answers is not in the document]
jumde: I'd be okay with that. cc: @flamsmark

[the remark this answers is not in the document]
flamsmark: I'd prefer "Allow $DOMAIN to open $APP?". I think that both the second-level domain and the app being opened are important. Do we have the ability to get the app's name?

FWIW. I tested on my phone and Chrome & DDG just open the link, Safari prompts with 'Open this page in "$APP"?', and Firefox does nothing.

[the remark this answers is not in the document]
jumde: I wanted to keep parity with the warning message on the desktop browser. I am not sure if there is an API to get the app name from domain(cc: @jhreis). I'll check.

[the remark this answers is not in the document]
diracdeltas: @flamsmark i think it's more important to show the URL of the page that is about to be opened in an external app than the URL of the embedding page (what you refer to as second-level domain). 

we show the embedder URL in desktop because there the embedder is granted a permission to open other external URLs in the future (which is bad UX that we are hopefully getting rid of soon); however that doesn't apply to mobile. on mobile, i don't see any reason to show the embedder URL. 

for now I think  "Allow $URL to be opened in an external application?" is sufficient, where `$URL` is the URL that is about to be opened.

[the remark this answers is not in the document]
jumde: screen shot 2018-03-08 at 3 45 05 pm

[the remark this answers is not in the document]
jumde: Done!

diracdeltas:
mockup lgtm, but defer to @jhreis

jamesmudgett:
Example standard-- where typically the action is in the title, and additional info in description.
ios11-iphone7-maps-allow-access-prompt

The two options here would be:

1. "Allow ... external application?" all in title value (vs description to appear bold.)
2. Add a title, something like "Allow link to switch apps?". Then change description read something like "'sms...' will launch external application". This also requires changing the buttons to read, "Allow", "Deny" vs "Open" and "Cancel"

Thanks!

jumde:
screen shot 2018-03-11 at 3 46 35 pm

cc: @jamesmudgett @jhreis

jumde:
screen shot 2018-03-11 at 3 50 21 pm

cc: @jamesmudgett @jhreis

jamesmudgett:
@jumde Thanks, looks great. Will defer to @jhreis for code review.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +172,18 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handleExternalURL (_ url: URL) -> Void {
+        let alertController = UIAlertController(title: "Allow link to switch apps?", message: " '" + url.relativeString + "' will launch an external application", preferredStyle: UIAlertControllerStyle.alert)
```
jhreis: Swift supports interpolation strings, a bit cleaner for things like this:
`"some '\(url.relativeString)'!!!"`

jhreis: All strings should be inside the `strings.swift` file (Client > Client > Frontend)

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +172,18 @@ extension BrowserViewController: WKCompatNavigationDelegate {
...
+        let cancelAction = UIAlertAction(title: "Don't Allow", style: UIAlertActionStyle.cancel) { (result : UIAlertAction) -> Void in
```
jhreis: Instead of using a return-only block, just add the explicit handler at the end of the action:
`...style: UIAlertActionStyle.cancel, handler: nil)` (no `->` or `{ return }` then

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +172,18 @@ extension BrowserViewController: WKCompatNavigationDelegate {
...
+            return
+        }
+        let openAction = UIAlertAction(title: "Allow", style: UIAlertActionStyle.default) { (result : UIAlertAction) -> Void in
```
jhreis: Remove type from `result`, and use `()` instead of `Void`
e.g.
`result -> ()`

brave/src/webview/BraveWebView.swift
```diff
@@ -659,11 +662,20 @@ extension BraveWebView: UIWebViewDelegate {
         }
     }
 
+    func isExternalUrl(_ url: URL) -> Bool {
+        return !inBrowserURISchemes.contains(url.scheme!)
```
jhreis: Should not use forced unwrappings (ending the `scheme` with a `!`). Either return an optional (Bool?), or more likely just return false under empty scheme:
`if let scheme = url.scheme { stuff } else { return false }`

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +172,18 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handleExternalURL (_ url: URL) -> Void {
```
jhreis: A more "Swift" name would be: 
`handle(externalURL: URL)` (do not need the trailing Void return / arrow)

brave/src/webview/BraveWebView.swift
```diff
@@ -659,11 +662,20 @@ extension BraveWebView: UIWebViewDelegate {
         }
     }
 
+    func isExternalUrl(_ url: URL) -> Bool {
```
jhreis: This would probably be a bit cleaner if we used an fileprivate extension on URL:
```
fileprivate extension URL {
    var isExternal: Bool {
        return // .contains(self)
    }
}
```

Then calling it would be as simple as `url.isExternal`

brave/src/webview/BraveWebView.swift
```diff
@@ -659,11 +662,20 @@ extension BraveWebView: UIWebViewDelegate {
...
+        return !inBrowserURISchemes.contains(url.scheme!)
+    }
+
     func webView(_ webView: UIWebView,shouldStartLoadWith request: URLRequest, navigationType: UIWebViewNavigationType ) -> Bool {
         guard let url = request.url else { return false }
 
         webView.backgroundColor = UIColor.white
 
+        if isExternalUrl(url) {
```
jhreis: Nice, not using parens around the statement!

brave/src/webview/BraveWebView.swift
```diff
@@ -659,11 +662,20 @@ extension BraveWebView: UIWebViewDelegate {
...
+            getApp().browserViewController.handleExternalURL(url)
```
jhreis: Very clean and code-simple solution.

brave/src/webview/BraveWebView.swift
jhreis: Wasn't entirely sure where to leave this comment, but I think _all_ of this PR should be relocated inside of `BrowserViewController+WKNavigationDelegate`, you will see relatively similar code inside of `webViewDecidePolicyForNavigationAction` where we handle other highly specific schemes. Your solution will work quite well with what is there, as those are handling specific schemes (e.g. tel), and urls (e.g. itunes, maps), and yours is a much more general solution that will work well for notification of 3rd party apps.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +189,16 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handle (_ externalURL: URL) -> Void {
```
jhreis: Remove the `_ ` to make the name explicit, instead of calling `handle(url), it will be `handle(externalURL: url)`

flamsmark:
I'd really prefer if it were possible to name the target application, as Safari does. Is that technically achievable?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +189,16 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handle (externalURL: URL) -> Void {
+        let alertController = UIAlertController(title: Strings.RequestSwitchAppsTitle, message: "'\(externalURL.relativeString)' \(Strings.RequestSwitchAppsMessage)", preferredStyle: UIAlertControllerStyle.alert)
+        let cancelAction = UIAlertAction(title: Strings.DontAllow, style: UIAlertActionStyle.cancel, handler: nil)
+        let openAction = UIAlertAction(title: Strings.Allow, style: UIAlertActionStyle.default) { result -> () in
```
jhreis: `result in` is enough, (remove `-> ()`)

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +189,16 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handle (externalURL: URL) -> Void {
```
jhreis: Remove trailing `-> Void` (no return type necessary)

Client/Frontend/Strings.swift
```diff
@@ -57,13 +57,17 @@ extension Strings {
...
     public static let NewFolder = NSLocalizedString("New Folder", comment: "Title for new folder popup")
     public static let EnterFolderName = NSLocalizedString("Enter folder name", comment: "Description for new folder popup")
     public static let Edit = NSLocalizedString("Edit", comment: "")
     
     public static let NewDevice = NSLocalizedString("Device Name", comment: "Title for new device popup")
     public static let DeviceFolderName = NSLocalizedString("Please enter a name for this device", comment: "Description for new device popup")
-    
+
+    public static let RequestSwitchAppsTitle = NSLocalizedString("Allow link to switch apps?", comment: "")
+    public static let RequestSwitchAppsMessage = NSLocalizedString("will launch an external application", comment: "")
```
jhreis: Should have caught this earlier, so in these strings files to signify runtime portions we add `%@`:
`"%@ will launch an external..."`
Then to use it we do:
`String(format: Strings.RequestSwitchAppsMessage, externalURL.relativeString)`
This provides some additional context to our translators, and makes it a bit more transparent.

Good example of this is `Open_All_Bookmarks` from the strings file.

jhreis:
@toml macOS has [Launch Services](https://developer.apple.com/documentation/coreservices/launch_services#//apple_ref/c/func/LSGetApplicationForURL), it however is not available on iOS. I do not believe this is currently possible with the user of public APIs. 😭 🤢

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -172,6 +189,16 @@ extension BrowserViewController: WKCompatNavigationDelegate {
         //history.setTopSitesNeedsInvalidation()
     }
 
+    func handle (externalURL: URL) {
+        let alertController = UIAlertController(title: Strings.RequestSwitchAppsTitle, message: String(format: Strings.RequestSwitchAppsMessage, externalURL.relativeString), preferredStyle: UIAlertControllerStyle.alert)
+        let cancelAction = UIAlertAction(title: Strings.DontAllow, style: UIAlertActionStyle.cancel, handler: nil)
+        let openAction = UIAlertAction(title: Strings.Allow, style: UIAlertActionStyle.default) { result -> () in
```
jhreis: Ditto on the `-> ()` inside the result block

flamsmark:
Alternative text proposal:

> **Open link in another app?**
> The site at $SCHEMEHOST wants to open $LINKURI in another app.
> [**Don't open**]
> [Open link]

Also: does this resist UI DOS: can a page just try to open a URI a bajillion times knowing that there's no way to navigate away while the prompt is open?

jhreis:
1. You mean splitting up the scheme v argument?:
brave://url=foo.bar
> The site at "brave" wants to open "foo.bar"
or like the requesting domain v the requesting redirect?:
> The site at "www.anotherdomain.com" wants to open "brave://foo.bar"

2. Interesting, may need to test this a bit more. This should only fire on a redirect, so if the user says "no" it should still attempt a navigation inside of Brave, which _should_ theoretically block additional requests. But I think we'll need to look at this a bit more.

That being said, there are currently prompts for items like phone calls and facetime, that we already prompt, and iOS has their native ones too, so this should not necessarily introduce new DOS avenues, although it may increase the motivation to do so (e.g. since people can prompt to direct to their own app)

flamsmark:
I mean calling out the site that's making the request and the request itself separately. So in [Pranjal's screenshot above](https://github.com/brave/browser-ios/pull/1478#issuecomment-372157080), it'd have been:

> **Open link in another app?**
> The site at http://yigitcanyilmaz.hol.es wants to open `sms:123456879&body=ThisIsProofofConcept` in another app.
> [**Don't open**]
> [Open link]

In that screenshot for instance, you can't see the whole of the requesting domain. So there's some risk of things like <trustedsite.org.snenanigans.malicious.biz> or what have you. I think that the two things that a user needs to make an informed choice are: 
1. who's asking to do this, and 
2. what are they asking to do. 

I'm not saying that everyone is going to read the prompt and think about it real hard, but the option should be there for those who want it.

* * * * *

I think that redirect to your own app is less annoying than constantly pushing the user to open an App Store link, and effectively denying access to the page (or maybe browser?) unless they do. It's an annoying advertiser practice, and I hope Shields would catch it, but better safe?

PR #1506, opened 2018-04-05
[WIP] Extended URP.

iccub:


brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -154,4 +169,30 @@ class UserReferralProgram {
...
+        service.fetchCustomHeaders() { headers, error in
+            if headers.isEmpty { return }
+
+            self.prefs.setObject(NSKeyedArchiver.archivedData(withRootObject: headers), forKey: CustomHeaderData.prefsKey)
+        }
+    }
+
+    class func addCustomHeaders(to request: URLRequest) -> URLRequest {
+
+        guard let prefs = BraveApp.getPrefs(), let customHeadersAsData: Data = prefs.objectForKey(CustomHeaderData.prefsKey) as Any? as? Data,
```
jhreis: `as Any? as? Data`?

iccub: `objectForKey` returns `T`, regular casting to Data throws `Generic parameter 'T' could not be inferred` so we need to end up with this ugly looking solution.

PR #1514, opened 2018-04-16
Bip-39 for sync, xcode 9.3 support for sync

iccub:
Switched Niceware with brave/crypto, creating new sync group on device uses bip-39 now.
Joining an existing group will detect whether 16 words niceware or 24 words bip-39 algorithm is being used.

Also included few fixes to make the code compile with Xcode 9.3, basically the same changes as on development branch.


Cartfile
```diff
@@ -13,6 +13,6 @@ github "brave/mixpanel-swift"     "limit_tracking"
 github "brave/sync" "staging"
 github "SwiftyJSON/SwiftyJSON"              "3.1.4"
 github "iziz/libPhoneNumber-iOS"            "0.9.2"
-github "diracdeltas/niceware" "v1.0.4"
 github "path/FastImageCache" "master"
 github "facebook/pop"
+github "brave/crypto" "master"
```
diracdeltas: you can also specify "0.1.0" instead of master. I've been publishing it to NPM: https://www.npmjs.com/package/brave-crypto

brave/src/sync/SyncCrypto.swift
```diff
@@ -51,7 +51,7 @@ class Niceware: JSInjector {
         // TODO: Add byteCount validation (e.g. must be even)
         
         executeBlockOnReady {
-            self.nicewareWebView.evaluateJavaScript("JSON.stringify(niceware.passphraseToBytes(niceware.generatePassphrase(\(byteCount))))") { (result, error) in
+            self.syncCryptoWebView.evaluateJavaScript("JSON.stringify(niceware.passphraseToBytes(niceware.generatePassphrase(\(byteCount))))") { (result, error) in
```
diracdeltas: is this supposed to still be calling niceware.generatePassphrase and niceware.passphraseToBytes? It looks like the result of this is just 32 random bytes, in which case you can just call `crypto.getSeed()` instead: https://github.com/brave/crypto/blob/master/docs/api.md#getSeed

brave/src/sync/SyncCrypto.swift
```diff
@@ -60,23 +60,24 @@ class Niceware: JSInjector {
...
     
     /// Takes hex values and returns associated English words
-    /// fromBytes: Array of hex strings (no "0x" prefix) : ["00", "ee", "4a", "42"]
-    /// returns (via completion): Array of words from niceware that map to those hex values : ["administrational", "experimental"]
+    /// fromBytesOrHex: Array of hex strings (no "0x" prefix) : ["00", "ee", "4a", "42"]
+    /// returns (via completion): String of words from brave-crypto that map to those hex values:
+    /// "administrational experimental"
     // TODO: Massage data a bit more for completion block
     func passphrase(fromBytes bytes: [Int], completion: @escaping (([String]?, Error?) -> Void)) {
+        let bipWordLenght = 24
```
diracdeltas: `Length` not `Lenght`. alternatively you could just use `crypto.passphrase.BIP39_32_BYTE_WORD_COUNT`: https://github.com/brave/crypto/blob/master/docs/api.md#passphrase.BIP39_32_BYTE_WORD_COUNT

brave/src/sync/SyncCrypto.swift
```diff
@@ -60,23 +60,24 @@ class Niceware: JSInjector {
...
                 (result, error) in
-                
-                let jsonArray = JSON(parseJSON: result as? String ?? "").array
-                let words = jsonArray?.map { $0.string }.flatMap { $0 }
-                
-                if words?.count != bytes.count / 2 {
+
+                // Words come in format: \"word1 word2 word3... \"
+                let words = (result as? String ?? "").replacingOccurrences(of: "\"", with: "").components(separatedBy: " ")
+                if words.count != bipWordLenght {
```
diracdeltas: Laptop supports both Niceware and Bip39 in this case, because there are existing sync users who are using Niceware. I don't know if this is a concern for iOS. If so then you  have to change this to `if words.count != bipWordLength && words.count != nicewareWordLength`

brave/src/sync/SyncCrypto.swift
```diff
@@ -136,26 +137,32 @@ class Niceware: JSInjector {
     }
     
     /// Takes English words and returns associated bytes (2 bytes per word)
```
diracdeltas: It's only two bytes per word for niceware. For bip39 I think it's 1.375 bytes per word.

brave/src/sync/SyncCrypto.swift
diracdeltas: `module.exports.passphrase.fromBytesOrHex` can take either niceware or bip39 words as input and returns the same output format so you don't have to worry about that. the only difference is whether the result is 16 or 24 words.

brave/src/sync/SyncCrypto.swift
```diff
@@ -136,26 +137,32 @@ class Niceware: JSInjector {
...
-    /// fromPassphrase: An array of words : ["administrational", "experimental"]
+    /// toBytes32: A string of words: "administrational experimental"
     /// returns (via completion): Array of integer values : [0x00, 0xee, 0x4a, 0x42]
     func bytes(fromPassphrase passphrase: Array<String>, completion: (([Int]?, Error?) -> Void)?) {
         // TODO: Add some keyword validation
+
+        let wordsWithSpaces = passphrase.joined(separator: " ")
         executeBlockOnReady {
+            let jsToExecute = "JSON.stringify(module.exports.passphrase.toBytes32(\"\(wordsWithSpaces)\"));"
```
diracdeltas: like https://github.com/brave/browser-ios/pull/1514/files#r182174600, this also wouldn't work for Niceware words (but I don't know if iOS has to support those). To support niceware, you'd have to do something like:

```
if passphrase.length === 16 {
  // Niceware
  let jsToExecute = "JSON.stringify(module.exports.passphrase.toBytes32(\"\(wordsWithSpaces)\", true));"
} else {
  let jsToExecute = "JSON.stringify(module.exports.passphrase.toBytes32(\"\(wordsWithSpaces)\"));"
}
```

brave/src/sync/SyncCrypto.swift
iccub: For creating a sync group we only do bip39 on iOS, for joining an existing sync group we determine which algo to use(bip or niceware).

brave/src/sync/SyncCrypto.swift
iccub: Doc for `toBytes32` says `Converts a 32-byte passphrase to uint8array bytes. Infers whether the passphrase is bip39 or niceware based on length.` Here https://github.com/brave/crypto/blob/ed4e252b54f04556620585c7e6e4d2d095245603/index.js#L201
So it will use niceware or bip, depending on how many words a user will type in.

diracdeltas: @iccub haha you're right, i forgot my own API

brave/src/sync/SyncCrypto.swift
diracdeltas: @iccub sgtm, i thought this code was also used for recovery.

brave/src/sync/SyncCrypto.swift
```diff
@@ -51,7 +51,7 @@ class Niceware: JSInjector {
         // TODO: Add byteCount validation (e.g. must be even)
         
         executeBlockOnReady {
-            self.nicewareWebView.evaluateJavaScript("JSON.stringify(niceware.passphraseToBytes(niceware.generatePassphrase(\(byteCount))))") { (result, error) in
+            self.syncCryptoWebView.evaluateJavaScript("JSON.stringify(niceware.passphraseToBytes(module.exports.getSeed(\(byteCount))))") { (result, error) in
```
jhreis: Don't think this JS should be `niceware` here

brave/tests_src/unit/NicewareTest.swift
```diff
@@ -8,7 +8,7 @@ import Shared
 class NicewareTest: XCTestCase {
```
jhreis: May as well just rename this (and vars). Otherwise will be confusing in future.

iccub: Renamed it, not sure why this comment didn't collapse.

iccub:
Related ticket: https://github.com/brave/browser-ios/issues/1527

PR #1532, opened 2018-04-19
Sync disabled for 1.6 release.

jhreis:


brave/src/frontend/BraveSettingsView.swift
```diff
@@ -102,11 +102,16 @@ class BraveSettingsView : AppSettingsTableViewController {
...
 
         settings += [
-            SettingSection(title: NSAttributedString(string: Strings.General.uppercased()), children: generalSettings),
+            SettingSection(title: NSAttributedString(string: Strings.General.uppercased()), children: generalSettings)
+        ]
             
-            SettingSection(title: NSAttributedString(string: Strings.OtherSettings.uppercased()), children:
-                [SyncDevicesSetting(settings: self)]
-            ),
+        #if !NO_SYNC
```
iccub: Double negative statements are unreadable, maybe change the flag to `SYNC_ENABLED` or add a helper method `isSyncEnabled` and check for `NO_SYNC` that flag there.

brave/src/sync/Sync.swift
```diff
@@ -143,6 +143,17 @@ class Sync: JSInjector {
     /// Notice:: seed will be ignored if the keychain already has one, a user must disconnect from existing sync group prior to joining a new one
     func initializeSync(seed: [Int]? = nil, deviceName: String? = nil) {
         
+        
+        #if NO_SYNC
+            if syncSeed == nil { return }
+            
+            leaveSyncGroup()
+            
+            let alert = UIAlertController(title: "Sync Disabled", message: "Sync has been disabled, as it will not be included in the next couple of production builds. Your iOS device has been auto-removed from any sync groups.", preferredStyle: .alert)
```
iccub: Line too long, split it.

brave/src/frontend/BraveSettingsView.swift
jhreis: Going to leave double negative, since I am eventually going to remove this flag, if we keep it "positive" then we will forever need to include that flag for sync to work. I do think we should do this for other things though, as most macros are defined as negatives.

jhreis: Kind of want to keep this on the compiler level. If I wrap this in a helper function it will be a runtime thing. Not a big deal. Once we release sync we will probably keep it around forever, so this code will almost certainly be removed soon.

PR #1550, opened 2018-04-26
Sync timestamp

jamesmudgett:


brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -4,12 +4,18 @@ import CoreData
 import SnapKit
 import Shared
 
-class MainSidePanelViewController : SidePanelBaseViewController {
+protocol MainSidePanelViewControllerDelegate {
+    func openSyncSetup() -> Void
```
iccub: `-> Void` can be removed

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -4,12 +4,18 @@ import CoreData
...
+}
+
+class MainSidePanelViewController : SidePanelBaseViewController, MainSidePanelViewControllerDelegate {
```
iccub: `:` should be next to `MainSidePanelViewController` without spaces between them.
Also putting `MainSidePanelViewControllerDelegate` in an extension is nice.

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -103,6 +119,21 @@ class MainSidePanelViewController : SidePanelBaseViewController {
         controller.modalPresentationStyle = UIModalPresentationStyle.formSheet
         present(controller, animated: true, completion: nil)
     }
+    
+    func openSyncSetup() {
+        if !Sync.shared.isInSyncGroup {
```
iccub: I would add an early return statement to avoid nesting like `if Sync.shared.isInSyncGroup { return }` but that's debatable

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -103,6 +119,21 @@ class MainSidePanelViewController : SidePanelBaseViewController {
...
+            syncSetupViewController = SyncWelcomeViewController()
+            syncSetupViewController!.navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(SEL_closeSync))
```
iccub: Unwrap it, add a line break and remove `SEL_` from selector

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -103,6 +119,21 @@ class MainSidePanelViewController : SidePanelBaseViewController {
...
+            
+            let settingsNavController = SettingsNavigationController(rootViewController: syncSetupViewController!)
```
iccub: Unwrap to remove `!`

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -103,6 +119,21 @@ class MainSidePanelViewController : SidePanelBaseViewController {
...
+            settingsNavController.modalPresentationStyle = UIModalPresentationStyle.formSheet
```
iccub: Remove `UIModalPresentationStyle`

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
```diff
@@ -103,6 +119,21 @@ class MainSidePanelViewController : SidePanelBaseViewController {
...
+            present(settingsNavController, animated: true, completion: nil)
```
iccub: `completion: nil` can be removed, it's nil by default.

brave/src/frontend/sync/SyncSettingsViewController.swift
```diff
@@ -125,7 +125,13 @@ class SyncSettingsViewController: AppSettingsTableViewController {
     }
     
     func SEL_done() {
-        navigationController?.popToRootViewController(animated: true)
+        guard let rootVC = navigationController?.viewControllers.first else { return }
```
iccub: We could unwrap it in a guard like this
```
guard let _ = navigationController?.viewControllers.first as? BraveSettingsView else { 
            dismiss(animated: true, completion: nil)
            return
        }
```

jhreis: I think we should actually pass in some sort of block for this. This controller should not have any concept how it is presented. Hard coding `BraveSettingsView` is very sensitive to any sort of changes. 

Having a `dismiss` closure, allows the presenter to give it a block to fire when it needs to go away. Much cleaner IMO.

Client.xcodeproj/project.pbxproj
```diff
@@ -1025,11 +1025,11 @@
 		0A3AA90B1FFF8C3100855082 /* FavoritesDataSource.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FavoritesDataSource.swift; sourceTree = "<group>"; };
 		0A3D9E64207FBC290034D5F0 /* crypto.js */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.javascript; name = crypto.js; path = Carthage/Checkouts/crypto/browser/crypto.js; sourceTree = SOURCE_ROOT; };
 		0A4932882019FC4800EB7DB8 /* ViewUtil.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ViewUtil.swift; sourceTree = "<group>"; };
```
fredbuluku: Thank for the guidance

fredbuluku:
Thank you to all who supported me in this conflict,be blessed.

brave/src/frontend/sidepanels/MainSidePanelViewController.swift
fredbuluku: Thanks for that,but there was a confusion in between that’s the reason I found myself in a mess but I believe the problem will be solved.

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -505,15 +508,57 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
     }
 
     override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
-        return nil
+        let label = UILabel()
+        label.font = UIFont.systemFont(ofSize: 14, weight: UIFontWeightRegular)
+        label.textAlignment = .center
+        
+        if !Sync.shared.isInSyncGroup {
```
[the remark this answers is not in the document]
fredbuluku: 🙏

brave/src/frontend/sync/SyncWelcomeViewController.swift
```diff
@@ -10,7 +10,10 @@ protocol NavigationPrevention {
 }
 
 class SyncWelcomeViewController: SyncViewController {
-
+    /// There is one use case when we show this VC in modal form, and after succesfully setting up sync
+    /// we need to dismiss the controller altogether.
+    var fromModal = false
```
jhreis: Instead of setting a flag like this, just set an actual `dismissHandler`, and then directly pass that into the settings one.

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -202,15 +204,18 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
         self.bookmarksFRC?.delegate = self
         
         NotificationCenter.default.addObserver(self, selector: #selector(self.reloadData), name: NotificationMainThreadContextSignificantlyChanged, object: nil)
+        
+        Sync.shared.syncFetchedHandlers.append { [weak self] in
```
jhreis: We should probably lock down `syncFetchedHandlers`, otherwise it could get abused pretty easily.

Probably best to make it private and create an `addSyncFetchedHandler`, which handles the appendage.

brave/src/sync/Sync.swift
```diff
@@ -608,7 +613,7 @@ extension Sync {
     func saveInitData(_ data: JSON) {
         // Sync Seed
         if let seedJSON = data["arg1"].array {
-            let seed = seedJSON.map({ $0.int }).flatMap({ $0 })
+            let seed = seedJSON.map({ $0.int }).compactMap({ $0 })
```
jhreis: Just use `compactMap` instead of both `map` and `compactMap` (I know this is existing code, but probably best to do anyway).

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -499,15 +502,57 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
...
 
     override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
-        return nil
+        let label = UILabel()
+        label.font = UIFont.systemFont(ofSize: 14, weight: UIFont.Weight.regular)
+        label.textAlignment = .center
+        
+        if Sync.shared.isInSyncGroup {
+            label.textColor = BraveUX.GreyG
+            label.numberOfLines = 1
```
jhreis: This is the default value. So should't need to be set.

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -165,8 +165,16 @@ class SyncDevicesSetting: Setting {
...
             navigationController?.pushViewController(syncSettingsView, animated: true)
+            
+            syncSettingsView.dismissHandler = {
+                navigationController?.popToRootViewController(animated: true)
+            }
+            
         } else {
             let view = SyncWelcomeViewController()
+            view.dismissHandler = {
+                view.navigationController?.popToRootViewController(animated: true)
```
jhreis: Isn't this presented in a modal? Or at least one of these? Previously we had one performing a `pop` and the other a `dismiss`.

iccub: This one is modal https://github.com/brave/browser-ios/pull/1550/files#diff-7ad63d4335f2f85bd7ebf4b45129bd8cR128

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -165,8 +165,16 @@ class SyncDevicesSetting: Setting {
             let syncSettingsView = SyncSettingsViewController(style: .grouped)
             syncSettingsView.profile = getApp().profile
             navigationController?.pushViewController(syncSettingsView, animated: true)
+            
+            syncSettingsView.dismissHandler = {
```
jhreis: Maybe we should move this before the `push` above to make it more uniform with the setup below.

iccub: Done. It didn't autoresolve this comment

PR #1560, opened 2018-05-02
fix #1556 - limit keyboard presentation when popup is presented -- sh…

### LINKED ISSUE #1556: Private browsing modal is blocked by keyboard when switching to PB mode
### Description:
Private browsing modal is blocked by keyboard when switching to PB mode

### Steps to Reproduce
  1. Clean install 1.6(18.04.30.15)
  2. Switch to PB mode
  3. After PB tab is opened, keyboard popsup covering the PB info modal

**Actual result:**
image
image

**Expected result:**
Info modal should be over the keyboard even if its shown.

100%

**Brave Version:**
1.6(18.04.30.15)

**Device details:**
All

**Website problems only:**

jamesmudgett:
…ould add to popupView to detect if non-subview has focus, only popup subviews should allow keyboard presentation


Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -39,8 +39,11 @@ extension BrowserViewController {
 
     func newPrivateTab() {
         openBlankNewTabAndFocus(isPrivate: true)
-        postAsyncToMain(0.2) {
-            self.selectLocationBar()
+        
+        if profile.prefs.boolForKey(kPrefKeySetBrowserLock) == true || profile.prefs.boolForKey(kPrefKeyPopupForBrowserLock) == true {
```
iccub: We do this check in two places. What about extracting it to a helper method to DRY?

iccub: We do this check in two places. What about extracting it to a helper method to DRY?

brave/src/frontend/PinController.swift
```diff
@@ -33,6 +33,14 @@ class PinViewController: UIViewController, PinViewControllerDelegate {
     var delegate: PinViewControllerDelegate?
     var pinView: PinLockView!
     
+    static var isBrowserLockEnabled: Bool {
+        if let profile = getApp().profile {
```
jhreis: Could reduce to:
```
let lock = profile.prefs.boolForKey(kPrefKeySetBrowserLock) == true || profile.prefs.boolForKey(kPrefKeyPopupForBrowserLock) == true
return lock ?? false
```

brave/src/frontend/BraveBrowserToolbarButtonActions.swift
```diff
@@ -22,7 +22,12 @@ extension BraveBrowserToolbarButtonActions {
     
     func respondToNewPrivateTab(action: UIAlertAction) {
         getApp().browserViewController.switchBrowsingMode(toPrivate: true)
-        let app = UIApplication.shared.delegate as! AppDelegate
-        app.browserViewController.urlBar.browserLocationViewDidTapLocation(app.browserViewController.urlBar.locationView)
+        
+        if PinViewController.isBrowserLockEnabled {
+            postAsyncToMain(0.2) {
```
jhreis: Why is this delay needed now?

jamesmudgett: Want to avoid a regression here if possible. This could have been an issue with animation or switching to PB which may have prevented input from being focused during transition.

jamesmudgett: Would require some testing.

PR #1562, opened 2018-05-02
Fix #1446, new folder crash regression

### LINKED ISSUE #1446: Creating new bookmark folder crashes brave
# Test plan
1. Attempt to create a bookmark folder on left wall
2. Folder should get created and browser should not crash 
3. Add a bookmark under folder, browser should not crash
4. Add a nested bookmark folder, browser should not crash

# original issue
2 closes issues found but seems like new bug.  Able to reproduced easily.  

### Description:

When creating a new bookmark folder brave crashes immediately 

### Steps to Reproduce

  1. Attempt to create new bookmark folder 
  2.
  3.

**Actual result:**

App crashes 

**Expected result:**

App should not crash and allow bookmark folder creation dialog 

**Brave Version:**

Version 1.5.3 (18.02.14.13)

**Device details:**

iPhone X 
11.3 beta 2
**Website problems only:**

iccub:
Autoscroll from here https://github.com/brave/browser-ios/pull/1553 caused this regression.
For some reason the app crashes when there are no bookmarks or folders added before adding a new folder.


jhreis:
It is because there is no item 0 in section 0.

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -342,7 +342,10 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
 
     func addFolder(titled title: String) {
         Bookmark.add(url: nil, title: nil, customTitle: title, parentFolder: currentFolder, isFolder: true)
-        tableView.scrollToRow(at: IndexPath(item: 0, section: 0), at: UITableViewScrollPosition.bottom, animated: true)
+        // The app crashes when scrolling to top when there are no bookmarks or folders added already
+        if let bookmarksCount = bookmarksFRC?.fetchedObjects?.count, bookmarksCount > 1 {
```
jhreis: I think using:
`if tableView.numberOfRows > 0` is probably safer.

Is it `> 1` (instead of `0`) just because there is no need to scroll?

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -342,7 +342,10 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
...
+            tableView.scrollToRow(at: IndexPath(item: 0, section: 0), at: .bottom, animated: true)
```
jhreis: @jamesmudgett should this be flipped to `.top` since folders are sorted towards the top?

Issue #1566, opened 2018-05-03
keychain items are exposed to icloud backups


diracdeltas:
from https://www.securetheorem.com/app/1/issues/006613:

```
The following items stored by the App in the Keychain are exported to backups:
sqlcipher.key.logins.db
pinLockInfo

Having sensitive data sent to iCloud exposes it to both Apple, Inc. and an attacker with the ability to compromise the user's iCloud account
```

Recommendation:
```
In order prevent Keychain data from being migrated to iTunes and iCloud backups, explicitly set a ThisDeviceOnly accessibility class (such as kSecAttrAccessibleWhenUnlockedThisDeviceOnly) for all Keychain items. More information about such accessibility classes is available at https://developer.apple.com/library/ios/documentation/security/Reference/keychainservices/Reference/reference.html.

Additionally, because the official iOS Keychain APIs (including SecItemAdd() and SecItemCopyMatching()) are overly complex and difficult to use, consider leveraging a wrapper instead in order to simplify the process of storing and retrieving data from the Keychain.

The Valet open source library can be used for this purpose: https://github.com/square/Valet.
```

diracdeltas:
not needed before migration

PR #1571, opened 2018-05-07
Disable iCloud/iTunes backups for keychain items

jumde:
Fixes: https://github.com/brave/browser-ios/issues/1566

When adding these items to the Keychain, sensitive data should be tagged with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, which prevents any sensitive data to be exported to iTunes and iCloud backups.

### DIFF
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -304,7 +304,7 @@ open class BrowserProfile: Profile {
         
         let Length: UInt = 256
         let secret = Bytes.generateRandomBytes(Length).base64EncodedString
-        self.keychain.set(secret, forKey: key)
+        self.keychain.set(secret, forKey: key, withAccessibility: .afterFirstUnlockThisDeviceOnly)
         return secret
     }()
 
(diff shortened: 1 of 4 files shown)


jhreis:
This is a tough one. It would be best to find ways to abstract this to guarantee these _must_ be on device, but I think the only real option is forking the dependency and adding it ourselves. I think it is probably worth doing at some point, but maybe too big of a change right now. 

So only real change here is there are still some missing instances. Do a search for `KeychainWrapper.standard.` and you should see a number of additional places.

Other potential issues:
- existing users will still be using the default keychain, do we want to migrate them to this?
- if not, we should not specify access level for read / removal as the keychains will be split between users.

jumde:
> It would be best to find ways to abstract this to guarantee these must be on device, but I think the only real option is forking the dependency and adding it ourselves.  I think it is probably worth doing at some point, but maybe too big of a change right now.

Agreed. May be we can just write a wrapper for now, and make sure that only the wrappers are used instead of the standard KeychainWrapper calls.

 > existing users will still be using the default keychain, do we want to migrate them to this?

Ah missed this, Thanks! I'll check if there is a way for us to find the existing items and update the attributes only. 

> if not, we should not specify access level for read / removal as the keychains will be split between users.

Agreed.

I think its better to wait for the outcome of https://github.com/brave/internal/issues/313 before making the decision.

jhreis:
Sounds good.

Yes, we could create a subclass of KeychainWrapper and just reimplement the methods we currently use (there aren't very many, maybe 6ish). That would be easier to keep track of than verifying the parameters are correct on any change. 

Good thoughts.

jumde:
Not required before migration.

PR #1574, opened 2018-05-08
fallback favicon, genrate from laptop's theme color script

jamesmudgett:
Also threw in new coinmarketcap icon (looks like they have a logo now)


Client.xcodeproj/project.pbxproj
```diff
@@ -3922,44 +3929,54 @@
 				TargetAttributes = {
 					288A2D851AB8B3260023ABC3 = {
 						CreatedOnToolsVersion = 6.2;
+						DevelopmentTeam = ULR7CC4V8L;
```
iccub: Remove dev teams.

Client/Assets/ThemeColor.js
```diff
@@ -0,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
+if (!window.__firefox__) {
+    window.__firefox__ = {};
+}
+
+window.__firefox__.sitecolor = function() {
```
iccub: Is it the same script browser laptop is using btw?

Client/Frontend/Browser/SiteColorManager.swift
```diff
@@ -0,0 +1,42 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import WebKit
+import Storage
+import WebImage
+
+
+class SiteColorManager : BrowserHelper {
```
iccub: Colon should be next to `SiteColorManager`

Client/Frontend/Browser/SiteColorManager.swift
```diff
@@ -0,0 +1,42 @@
...
+        return "siteColorMessageHandler"
+    }
+    
+    func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage) {
+        guard let tab = browser else { return }
+        guard let data = message.body as? [String: String], let themeColor = data["themeColor"] else { return }
+        
+        // Convert rgb(#, #, #) to UIColor
+        let rgbRaw: [Substring] = themeColor.split(separator: "(")[1].split(separator: ")")[0].split(separator: " ").joined(separator: "").split(separator: ",")
+        let rgb: [Int] = rgbRaw.map { Int(String($0))! }
```
iccub: I would avoid force unwrap here, we don't want random crashes.

brave/src/frontend/favorites/FavoritesHelper.swift
```diff
@@ -45,11 +45,31 @@ struct FavoritesHelper {
...
+            let rectangle = CGRect(x: 0, y: 0, width: FallbackIconUX.size.width, height: FallbackIconUX.size.height)
+            
+            ctx.cgContext.addRect(rectangle)
+            ctx.cgContext.setFillColor(color.cgColor)
+            ctx.cgContext.drawPath(using: .fillStroke)
+            
+            let paragraphStyle = NSMutableParagraphStyle()
+            paragraphStyle.alignment = .center
+            
+            let attrs = [NSFontAttributeName: UIFont(name: "HelveticaNeue-Thin", size: iconSize.height-90)!, NSParagraphStyleAttributeName: paragraphStyle, NSForegroundColorAttributeName: BraveUX.White]
```
iccub: What about removing `!` with a fallback to use system font? I know helvetica is bundled by apple but who knows if they are going to change it

brave/src/frontend/favorites/FavoritesHelper.swift
```diff
@@ -45,11 +45,31 @@ struct FavoritesHelper {
...
+            
+            let string: NSString = NSString(string: "\(letter)".uppercased())
```
iccub: `letter` is String no need to wrap it in `\()`

brave/src/frontend/favorites/FavoritesTileDecorator.swift
```diff
@@ -109,13 +117,32 @@ class FavoritesTileDecorator {
...
-                            return
+                        var final: UIImage? = img
+                        var useFallback: Bool = false
+
+                        if img == nil {
+                            useFallback = true
+                        } else if let img = img, img.size.width < FallbackIconUX.minSize.width && img.size.height < FallbackIconUX.minSize.height {
+                            useFallback = true
+                        } else if let img = img {
+                            debugPrint("Favicon size: \(img.size)")
```
iccub: replace `debugPrint` with `log`

brave/src/frontend/favorites/FavoritesTileDecorator.swift
```diff
@@ -109,13 +117,32 @@ class FavoritesTileDecorator {
             else {
                 postAsyncToMain {
                     cell.imageView.sd_setImage(with: iconUrl, completed: { (img, err, type, url) in
-                        guard let img = img else {
-                            // avoid retrying to find an icon when none can be found, hack skips FaviconFetch
-                            ImageCache.shared.cache(ThumbnailCellUX.PlaceholderImage!, url: cacheWithUrl, type: .square, callback: nil)
-                            cell.imageView.image = ThumbnailCellUX.PlaceholderImage
-                            return
+                        var final: UIImage? = img
```
iccub: `: UIImage?` can be removed. Also maybe renaming `final` to something better? It even gets recognized as a keyword in githubs code review and can get confusing

brave/src/frontend/favorites/FavoritesTileDecorator.swift
```diff
@@ -109,13 +117,32 @@ class FavoritesTileDecorator {
...
+                        var useFallback: Bool = false
```
iccub: `: Bool` can be removed

brave/src/frontend/favorites/FavoritesTileDecorator.swift
```diff
@@ -109,13 +117,32 @@ class FavoritesTileDecorator {
...
+
+                        if img == nil {
+                            useFallback = true
+                        } else if let img = img, img.size.width < FallbackIconUX.minSize.width && img.size.height < FallbackIconUX.minSize.height {
+                            useFallback = true
+                        } else if let img = img {
+                            debugPrint("Favicon size: \(img.size)")
+                        }
+
+                        if useFallback, let host = self.url.host, let letter = host.replacingOccurrences(of: "www.", with: "").first {
```
iccub: You are stripping off `www` but what about other subdomains like `admin.webapp.com`?

brave/src/frontend/favorites/FavoritesTileDecorator.swift
```diff
@@ -109,13 +117,32 @@ class FavoritesTileDecorator {
...
+                            var tabColor = FallbackIconUX.color
+                            
+                            // Only use stored color if it's not too light.
+                            if let color = self.color, !color.isLight {
+                                tabColor = color
+                            }
+                            
+                            final = FavoritesHelper.fallbackIcon(withLetter: String(letter), color: tabColor, andSize: FallbackIconUX.size)
```
iccub: `fallbackIcon` is used only here, maybe remove `andSize` param and put it inside of the method, `FallbackIconUX.size` is used there already.

Client/Frontend/Browser/Browser.swift
```diff
@@ -104,6 +104,9 @@ class Browser: NSObject, BrowserWebViewDelegate {
     var restoring: Bool = false
     var pendingScreenshot = false
     
+    // Tab color
+    var color: UIColor?
```
iccub: Maybe `siteColor, bgColor, faviconColor` would be better? Not sure if we are planning to use this attribute for other features than the favicon fallback.
That would require renaming it in other places too

brave/src/frontend/favorites/FavoritesTileDecorator.swift
jamesmudgett: Others would be ok, eg. foodie.tumblr.com would display 'F'

Client/Frontend/Browser/Browser.swift
jamesmudgett: Thinking in the future we could style the toolbars or tab, similar to desktop.

Client/Assets/ThemeColor.js
jamesmudgett: Same script, has a few modifications:
1. We grab color samples from y=1 vs y=3
2. We check for meta theme color first.

brave/src/frontend/favorites/FavoritesHelper.swift
jamesmudgett: good call

Client/Frontend/Browser/SiteColorManager.swift
```diff
@@ -0,0 +1,49 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import WebKit
+import Storage
+import WebImage
+
+
+class SiteColorManager: BrowserHelper {
+    let profile: Profile!
```
jhreis: No need to force unwrap this if set in the init. This removes the compiler checks to verify the object is actually present.

brave/src/frontend/favorites/FavoritesHelper.swift
```diff
@@ -45,11 +45,31 @@ struct FavoritesHelper {
...
+            let rectangle = CGRect(x: 0, y: 0, width: iconSize.width, height: iconSize.height)
+            
+            ctx.cgContext.addRect(rectangle)
+            ctx.cgContext.setFillColor(color.cgColor)
+            ctx.cgContext.drawPath(using: .fillStroke)
+            
+            let paragraphStyle = NSMutableParagraphStyle()
+            paragraphStyle.alignment = .center
+            
+            let attrs = [NSFontAttributeName: UIFont(name: "HelveticaNeue-Thin", size: iconSize.height-90) ?? UIFont.systemFont(ofSize: iconSize.height-90, weight: UIFontWeightThin), NSParagraphStyleAttributeName: paragraphStyle, NSForegroundColorAttributeName: BraveUX.White]
```
iccub: Break this line

Client/Frontend/Browser/SiteColorManager.swift
jamesmudgett: Cool, will fix. Just mirrored the others.

PR #1578, opened 2018-05-09
Fix #1561, fix #739 authentication for local webserver requests

### LINKED ISSUE #739: Dev issue: GCDWebserver uuid url param (merge from fx ios)
WebServer.swift should grab the latest changes from fx iOS, particular the guard against external apps calling WebServer localhost URLs.
More info:
https://github.com/mozilla-mobile/firefox-ios/issues/2474

### LINKED ISSUE #1561: Add authentication to requests to local webserver
requests to the local server (https://github.com/swisspol/GCDWebServer) should be rejected unless they are authenticated to be from the Brave app (versus other apps on the same device). 

Proposal:
1. iOS generates a secret token at startup; for instance a 32-byte random number
2. token is shared between GDCWebServer and the context that is making requests to GDCWebServer (not sure how this is actually done)
3. When the local server gets the request, it verifies that the token is correct. If so then it completes the request. If not then it returns 401.

iccub:
`NSUUID` is used for generating a random token. More info here https://developer.apple.com/documentation/foundation/nsuuid

Removed a handler for license request, as we don't use it.

FYI: Local web server is used for: 
- opening websites in reader mode
- showing error pages
- restoring session
- showing home panel(I'm not sure about this as we don't serve any html, just showing regular iOS view controller, there is a handler for `about/home` but it never gets called)

After migrating to WKWebview, adding auth will a lot of easier, this PR is more 'hacky'


iccub:
Closing, will resubmit another solution

Client/Frontend/Browser/ErrorPageHelper.swift
```diff
@@ -133,22 +133,24 @@ class ErrorPageHelper {
         self.certStore = certStore
 
         server.registerHandlerForMethod("GET", module: "errors", resource: "error.html", handler: { (request) -> GCDWebServerResponse! in
-            guard let url = ErrorPageHelper.originalURLFromQuery((request?.url)!) else {
+            guard let url = ErrorPageHelper.originalURLFromQuery((request?.url)!), let request = request else {
```
jhreis: unwrap `request`, then use that inside the `url` unwrapping.

Client/Application/WebServer.swift
```diff
@@ -72,4 +75,18 @@ class WebServer {
...
+        var request = URLRequest(url: url)
+        request.addValue(uniqueBytes, forHTTPHeaderField: headerAuthKey)
+        
+        return request
+    }
+    
+    func isRequestAuthenticated(_ request: GCDWebServerRequest) -> Bool {
+        guard let headerValue = request.headers[WebServer.headerAuthKey] as? String else { return false }
+        
+        return headerValue == WebServer.uniqueBytes
```
diracdeltas: since `==` aborts as soon as there is a byte mismatch, the attacker could try different prefixes and measure the time it takes for the function to run until they have the entire value. this is usually solved by using a constant-time comparison function (see https://gist.github.com/levigross/8691015 for some examples)

another way to solve the issue is just to compare the hash of the two values instead of the values themselves, ex `sha256(headerValue) == sha256(WebServer.uniqueBytes)`

iccub: That's interesting, thanks. I went with sha256 approach as we have implemented it already.

Client/Frontend/Browser/AboutHomeHandler.swift
```diff
@@ -9,22 +9,7 @@ import GCDWebServers
...
-    static func register(_ webServer: WebServer) {
-        webServer.registerHandlerForMethod("GET", module: "about", resource: "license") { (request: GCDWebServerRequest!) -> GCDWebServerResponse! in
-            let path = Bundle.main.path(forResource: "Licenses", ofType: "html")
-            do {
-                let html = try NSString(contentsOfFile: path!, encoding: String.Encoding.utf8.rawValue) as String
-                return GCDWebServerDataResponse(html: html)
-            } catch {
-                print("Unable to register webserver \(error)")
-            }
+        webServer.registerHandlerForMethod("GET", module: "about", resource: "home") { (request: GCDWebServerRequest!) -> GCDWebServerResponse? in
```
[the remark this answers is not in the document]
iccub: `/about/home/` handler never gets called for some reason. Going to add auth check to be extra safe.
Also added one for session restore

diracdeltas:
lgtm now

PR #1588, opened 2018-05-14
Convert https error from popups to prints.

iccub:
Motivation: these popups(especially the one after 10 sec delay) happen on every launch and it's tiring to dismiss them every time.
It's only used in debug so we can safely use `print()` instead.


jhreis:
I think this was for QA to see. Any idea why these are hit every time?

PR #1594, opened 2018-05-15
Coredata context improvements

jhreis:
This is a stepping stone in utilizing the newer `NSPersistentContainer`.


jhreis:
This work is important, but will not be done on this code base, so going to close this out. Good for reference if needed though.

PR #1599, opened 2018-05-16
Fix #1596, enable all regional blocklists.

### LINKED ISSUE #1596: Enable All Regional Blocklists by Default
Android has been doing this since last release without issue. iOS should do the same. We currently are already auto-enabled for: `["ru", "uk", "be", "hi", "sv"]`, but should just expand this globally.

See Adblock.swift for implementation.

iccub:


brave/src/webfilters/AdBlocker.swift
```diff
@@ -29,8 +29,6 @@ class AdBlocker {
         }
     }
     fileprivate var isRegionalAdblockEnabled: Bool? = nil
```
jhreis: Can we just make this a non-optional now?

brave/src/webfilters/AdBlocker.swift
```diff
@@ -28,11 +28,12 @@ class AdBlocker {
             updateRegionalAdblockEnabledState()
         }
     }
-    fileprivate var isRegionalAdblockEnabled: Bool? = nil
+    fileprivate let isRegionalAdblockEnabled: Bool
 
     fileprivate init() {
+        isRegionalAdblockEnabled = BraveApp.getPrefs()?.boolForKey(AdBlocker.prefKeyUseRegional) ?? true
```
jhreis: Instead of just using `true` here we should use `prefKeyUseRegionalDefaultValue` which is already set to `true.

brave/src/webfilters/AdBlocker.swift
```diff
@@ -81,21 +79,17 @@ class AdBlocker {
 
     func isRegionalAdblockPossible() -> (hasRegionalFile: Bool, isDefaultSettingOn: Bool) {
         return (hasRegionalFile: currentLocaleCode != "en" && regionToS3FileName[currentLocaleCode] != nil,
-                isDefaultSettingOn: isRegionalAdblockEnabled ?? false)
+                isDefaultSettingOn: isRegionalAdblockEnabled)
    }
 
     func updateEnabledState() {
-        isNSPrefEnabled = BraveApp.getPrefs()?.boolForKey(AdBlocker.prefKey) ?? AdBlocker.prefKeyDefaultValue
+        isNSPrefEnabled = BraveApp.getPrefs()?.boolForKey(AdBlocker.prefKey) ?? AdBlocker.prefKeyUseRegionalDefaultValue
```
jhreis: This shouldn't be changed, this is for general adblocking, and not regional.

PR #1608, opened 2018-05-18
Issue 1461 Double Status Bar Side Panel Fix

nickdobos:
Starting small with my first ever open source contribution! I think I've got the fork PR setup correctly, let me know if I need to change anything about that. 

Fixes: https://github.com/brave/browser-ios/issues/1461

Fixed a layout issue that would occur when SidePanelBaseVC's containe…rView got de-synced from the scrollView. This could happen in the event of the view resizing without transitioning size classes. Such as when a double status bar is displayed.

may-17-2018 21-53-24


jamesmudgett:
Fantastic fix, had been on my list!

PR #1636, opened 2018-06-12
Sync reordering

iccub:
Notes:
- copied bookmark_util.js from crypto lib to a new file, to utilize synchronous javascript calls. It made reordering work easier to do.
- sync reordering algorithms is now used even if not in a sync group. The exception is favorites, where old `order` attribute is being used.
- removing `syncUUID` while joining a sync group is not implemented yet
- there is an update bug, I want to fix it on a separate branch(it was introduced before my changes)


brave/src/sync/Sync.swift
```diff
@@ -570,9 +609,32 @@ extension Sync {
...
 
+    }
+    
+    func getBookmarkOrder(previousOrder: String?, nextOrder: String?) -> String? {
+        
+        // Empty string as a parameter means next/previous bookmark doesn't exist
+        let prev = previousOrder ?? ""
+        let next = nextOrder ?? ""
+        
+        let function = jsContext?.objectForKeyedSubscript("getBookmarkOrder")
```
jhreis: `function` should be more descriptive.

brave/src/sync/Sync.swift
```diff
@@ -570,9 +609,32 @@ extension Sync {
...
+        return function?.call(withArguments: [prev, next]).toString()
+    }
+    
+    fileprivate func setBaseBookmarkOrder() {
+        guard let deviceId = Device.currentDevice()?.deviceId?.first else { return }
+        let function = jsContext?.objectForKeyedSubscript("getBaseBookmarksOrder")
```
jhreis: `function` should be more descriptive.

brave/src/sync/Sync.swift
```diff
@@ -355,6 +389,11 @@ extension Sync {
             return
         }
         
+        if let syncBookmark = records.first?.recordType, syncBookmark == .bookmark, baseSyncOrder == nil {
```
jhreis: Since equality can just do:

`records.first?.recordType == .bookmark && baseSyncOrder == nil {}`

jhreis: The check here can be removed, as this actually automatically handles this: https://github.com/brave/browser-ios/pull/1636/files#diff-d4a71e14f4d1211c6b1ea257fdbe5e94R382

brave/src/sync/Sync.swift
```diff
@@ -121,6 +132,29 @@ class Sync: JSInjector {
...
+        context?.exceptionHandler = { _, exc in
+            log.error(exc.debugDescription)
+        }
+        guard let path = Bundle.main.path(forResource: "bookmark_util", ofType: "js") else {
+            log.error("Could not load bookmark_util.js")
+            return nil
+        }
+        
+        do {
+            let scriptAsString = try String(contentsOfFile: path, encoding: String.Encoding.utf8)
```
jhreis: Leave some comments in here if possible, a bit weird seeing the JS loaded into a context, and result ignored (I know it is to leverage the usage later).

brave/src/sync/Sync.swift
```diff
@@ -608,6 +671,7 @@ extension Sync: WKScriptMessageHandler {
             print("---- Sync Debug: \(data)")
         case "sync-ready":
             self.isSyncFullyInitialized.syncReady = true
+            setBaseBookmarkOrder()
```
jhreis: Any reason this isn't inside `checkIsSyncReady`?

brave/src/sync/Sync.swift
```diff
@@ -599,6 +661,7 @@ extension Sync: WKScriptMessageHandler {
             // (e.g. arg2 is [Int])
             let data = JSON(parseJSON: message.body as? String ?? "")
             self.saveInitData(data)
+            Bookmark.removeSyncOrders()
```
jhreis: Not understanding this one, can you explain?

brave/src/data/Bookmark.swift
```diff
@@ -257,30 +282,62 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
-            bms.insert(src, at: bms.index(of: dest)!)
-        } else {
-            let end = bms.index(of: dest)! + 1
-            bms.insert(src, at: end)
-        }
-        
-        for i in 0..<bms.count {
-            bms[i].order = Int16(i)
+        // Favorites use regular order algorithm
+        if src.isFavorite {
```
jhreis: These need to be split up, this function is massive.

brave/src/sync/Sync.swift
iccub: `syncOrder` is now used internally for ordering even if user is not using sync. So bookmarks go like `0.0.1, 0.0.2...` after we join a sync group we are removing these sync order and setting them again, now with a proper device id, so we can replace `0.x.x` with `device_id.x.x`

jhreis: But this will be called every time sync is initialized, right?

iccub: Yes, but sync is initialized only once. Is it a problem?
New sync join = new device id, we need to update bookmarks anyway

jhreis: Ah, I thought it was called every time the library was initialized, as in, every launch. Would add comment here, but lgtm.

brave/src/data/Bookmark.swift
```diff
@@ -219,7 +230,21 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
         
         return bk
     }
-  
+    
+    class func allBookmarksOfAGivenLevelPredicate(parent: Bookmark?) -> NSPredicate {
+        let isFavoriteKP = #keyPath(Bookmark.isFavorite)
+        let parentFolderKP = #keyPath(Bookmark.parentFolder)
+        
+        if let parentFolder = parent {
```
jhreis: Can this be reduced to:
`return NSPredicate(format: "%K == %@ AND %K == NO", parentFolderKP, parent ?? "nil", isFavoriteKP)`

brave/src/data/Bookmark.swift
```diff
@@ -257,30 +282,62 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
+            let isMovingUp = sourceIndexPath.row > destinationIndexPath.row
+            
+            // Depending on drag direction, all other bookmarks are pushed up or down. 
+            if isMovingUp {
+                var prev: String?
+                
+                // Bookmark at the top has no previous bookmark.
+                if destinationIndexPath.row > 0 {
+                    let index = IndexPath(row: destinationIndexPath.row - 1, section: destinationIndexPath.section)
+                    prev = (frc.object(at: index) as! Bookmark).syncOrder
```
jhreis: We should remove all of these forced castings. Change https://github.com/brave/browser-ios/pull/1636/files#diff-3ec30f0d026baec8beed697836bb1b02R276

to:
`guard let frc = frc as? NSFetchedResultsController<Bookmark> else { return }`

This allows `frc` to always return `Bookmark`

brave/src/sync/Sync.swift
iccub: This is causing race conditions as well, going to move it to `checkIsSyncReady`.

brave/src/sync/Sync.swift
iccub: I added a comment one line below :) just fyi bc github didn't collapse this comment.

brave/src/sync/bookmark_util.js
```diff
@@ -0,0 +1,125 @@
+'use strict'
+
+
```
iccub: I'm thinking about adding a comment here, to not forget that this is our handmade converted file from `brave/sync`

brave/src/sync/Sync.swift
```diff
@@ -355,6 +410,12 @@ extension Sync {
             return
         }
         
+        if recordType == .bookmark && baseSyncOrder == nil {
```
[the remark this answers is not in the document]
iccub: I'm not sure if I get it.

These are separate checks,
first check unwraps record type, and second check only happens if record type is a bookmark.
Recordtype can be a device, preferences etc. so this second check is only for sync record types.

For example if we are sending `.device` type of record type, we do not care about `baseSyncOrder`

PR #1652, opened 2018-06-22
Fix #1649, disable ddg popup is it was chosen already.

### LINKED ISSUE #1649: DDG modal is still displayed for Private Tabs if you've already made the selection in Settings
### Description:
If you go into settings and set your search to DDG for a private tab and then go to Private mode for the first time, you still get the modal asking if you want to 'Private search with DuckDuckGo?'.

### Steps to Reproduce

  1. Clean install of 1.6.2 (18.06.20.18)
  2. Navigate to settings
  3. Set search engine for Private tabs to DDG
  4. Open a Private tab

**Actual result:**
You get the 'Private search with DuckDuckGo?' modal.

**Expected result:**
No modal since you've already consciously made the decision to change your search engine prior to entering private mode

Easily

**Brave Version:**
1.6.2 (18.06.20.18)

**Device details:**
all

**Website problems only:**

### Additional Information
Found by @kjozwiak

iccub:


Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -606,6 +606,11 @@ class BrowserViewController: UIViewController {
             return
         }
         
+        // Do not show ddg popup if user already chose it for private browsing. 
+        if profile?.searchEngines.defaultEngine(forType: .privateMode).shortName == "DuckDuckGo" {
```
jhreis: Should should make "DuckDuckGo" a constant. I see it used elsewhere, so we should find a central place to put it.

PR #1716, opened 2018-08-27
Search engines update

iccub:


Client/Frontend/Browser/SearchEngines.swift
```diff
@@ -53,6 +55,22 @@ class SearchEngines {
...
     }
     
+    private func initSearchEnginesForRegion() {
+        // Never override existing user's search engine preferences. 
+        let isFirstLaunch = prefs.arrayForKey(DAU.preferencesKey) == nil
+        if !isFirstLaunch { return }
+        
+        guard let region = Locale.current.regionCode else { return }
+        
+        switch region {
```
jhreis: Using the static `defaultRegionSearchEngines` above, this would be converted to something like:

```
guard let searchEngine = defaultRegionSearchEngines[region] else { return }

defaultEngine(OpenSearchEngine.EngineNames.qwant, forType: .standard)
defaultEngine(OpenSearchEngine.EngineNames.qwant, forType: .privateMode)
```

Client/Frontend/Browser/BrowserViewController/BrowserViewController.swift
```diff
@@ -602,8 +602,9 @@ class BrowserViewController: UIViewController {
     
     var shouldShowDDGPromo: Bool {
         // We want to show ddg promo in most cases so guard returns true.
-        guard let region = Locale.current.regionCode else { return true } 
-        return region != "DE" && region != "FR" 
+        guard let region = Locale.current.regionCode else { return true }
+        
+        return SearchEngines.defaultRegionSearchEngines[region] != OpenSearchEngine.EngineNames.qwant
```
jhreis: Could probably just return false if there is any value at all. I think this being set to anything by default will probably override the DDG stuff.

PR #1730, opened 2018-09-04
Regional search delegate fix

iccub:
We didn't check for current region while setting the regional search preference.

### DIFF
--- a/Client/Frontend/Browser/SearchEngines.swift
+++ b/Client/Frontend/Browser/SearchEngines.swift
@@ -81,6 +81,16 @@ class SearchEngines {
         }
     }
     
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/SearchEngines.swift
```diff
@@ -81,6 +81,16 @@ class SearchEngines {
         }
     }
     
+    class func regionalSearchEnginesFlagsSetup(prefs: Prefs) {
+        guard let region = Locale.current.regionCode,
```
jhreis: Great for now, will need to be simplified a bit in the future possibly.

PR #1732, opened 2018-09-11
Swift 4 migration.

iccub:
This PR is too big to for code review alone, we need to do a full qa run checking if the app generally works.

### Overview

- In addition to code changes to make it work with swift 4, `URLConnection` needed to be replaced to make reader mode work.
The commit for it can be found here https://github.com/brave/browser-ios/pull/1732/commits/6f240f8d9ba9bb041f59cf8ed9e13fad6c1c3cff and should be code reviewed.

- The app compiles on Xcode 9(9F2000) and Xcode 10 GM(10A254a)

- `brave/crypto` lib had to be updated to `0.2.0` due to build errors.

- `mixpanel` was removed altogether, we were not using it anymore and the lib wasn't compiling with swift4. Only thing left is unused `MIXPANEL_TOKEN` in `brave.plist`

- `libstdc++` was removed, it goes away in iOS 12, no replacement was needed for it, the app runs fine without it

- We do not use carthage for `sqlite.swift`, a newer version was copied manually, checking diffs of sqlite classes isn't needed.

I did some testing on simulator and my iPhone 7 iOS 11 and noticed no crashes/weird behaviors.


Client/Frontend/Widgets/TabsButton.swift
```diff
@@ -144,6 +144,8 @@ class TabsButton: UIControl {
         fatalError("init(coder:) has not been implemented")
     }
 
+    // fix swift4
```
jhreis: ?

Shared/RollingFileLogger.swift
```diff
@@ -22,7 +21,7 @@ open class RollingFileLogger: XCGLogger {
 
     let root: String
 
-    public init(filenameRoot: String, logDirectoryPath: String?, sizeLimit: Int64 = TwoMBsInBytes) {
+    public init(filenameRoot: String, logDirectoryPath: String?, sizeLimit: Int64 = 2 * 100000) {
```
jhreis: Why remove this constant? Makes it much less clear what the `sizeLimit` value is.

Utils/DynamicFontHelper.swift
```diff
@@ -48,25 +48,25 @@ class DynamicFontHelper: NSObject {
         return deviceFontSize
     }
     var DeviceFont: UIFont {
-        return UIFont.systemFont(ofSize: deviceFontSize, weight: UIFontWeightMedium)
+        return UIFont.systemFont(ofSize: deviceFontSize, weight: UIFont.Weight.medium)
```
jhreis: This was probably automated, see this everywhere. We can just use `.medium` instead of the entire `UIFont.Weight.medium`, but probably not worth actually changing rn.

brave/src/BraveApp.swift
```diff
@@ -94,53 +90,13 @@ class BraveApp {
     }
 
     class func didFinishLaunching() {
-        #if !NO_FABRIC
```
jhreis: Mixpanel removal is good, but really probably shouldn't have been included in this PR.

brave/src/frontend/sync/SyncAddDeviceTypeViewController.swift
```diff
@@ -0,0 +1,139 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
jhreis: Where did this file come from?

brave/src/frontend/sync/SyncBarcodeView.swift
```diff
@@ -0,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
jhreis: Ditto, on file creation?

brave/src/webfilters/UrlProtocol.swift
```diff
@@ -15,9 +17,18 @@ let markerRequestHandled = "request-already-handled"
 
 class URLProtocol: Foundation.URLProtocol {
 
-    var connection: NSURLConnection?
+    var session: URLSession?
```
jhreis: Very nice job.

brave/src/webview/BraveWebView.swift
```diff
@@ -619,18 +619,18 @@ class BraveWebView: UIWebView {
             recentlyBlocked = RecentlyBlocked()
         case .httpseIncrement:
             shieldStats.httpse += increment
-            BraveGlobalShieldStats.singleton.httpse += increment
+            BraveGlobalShieldStats.singleton.httpse = BraveGlobalShieldStats.singleton.httpse.advanced(by: increment)
```
jhreis: This is much more error prone, since it requires the correct name on both sides.

We should either:
- cast `increment` to an `Int32` so it can be incremented via `+=`
- adjust `singleton.httpse` to be a standard `Int`
- change all the other `Int`s here to be `Int32` (e.g. `shieldStats.httpse` and also the parameter `increment`

brave/src/BraveApp.swift
iccub: It did not compile otherwise, I would need to upgrade mixpanel dependency which is also out of scope for this PR

Utils/DynamicFontHelper.swift
iccub: This was an Xcode's suggested replacement, I I can fix it quickly with find/replace.

jhreis: Don't bother IMO.

brave/src/BraveApp.swift
jhreis: Sounds good, yeah, ignore mixpanel stuff beyond this rn, I actually have some work done on this too, so will submit PR for that once this is merged.

brave/src/frontend/sync/SyncAddDeviceTypeViewController.swift
iccub: Good catch. I left them by mistake while rebasing my old swift4 migration branch. They were renamed and I didn't delete them completely.

This was renamed to `SyncSelectDeviceTypeViewController`

brave/src/frontend/sync/SyncBarcodeView.swift
iccub: This was renamed to `SyncQRCodeView`, left it by mistake while rebasing sorry.

Shared/RollingFileLogger.swift
iccub: Compiler complained about it, I will fix it by adding `public` to the constant

PR #1734, opened 2018-09-26
Using URP staging URL for beta/development builds.

jhreis:


brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -22,12 +22,16 @@ class UserReferralProgram {
...
 
-        guard let host = getPlistString(for: UserReferralProgram.hostPlistKey),
+        guard var host = getPlistString(for: UserReferralProgram.hostPlistKey),
             let apiKey = getPlistString(for: UserReferralProgram.apiKeyPlistKey), let prefs = getApp().profile?.prefs else {
                 log.error("Urp init error, failed to get values from Brave.plist.")
                 return nil
         }
 
+        if kIsDevelomentBuild {
+            host = UserReferralProgram.stagingUrl
```
iccub: Now we have staging url in two places, in `URP_HOST` .plist and in `UserReferralProgram.stagingUrl`.
And if for some reason we have a development build and the URP_HOST plist is not set, it will go to error on line 27.

I would extract `host` variable from line 25 and put it in a separate if-else clause.

So it would be:
if development => use staging url
else if plist string is not nil => use plist setting
else => log error

jhreis: Gotcha, I added this since the Plist `host` was never being set correctly, probably something with the script. Will take your approach!

jhreis:
Closing in favor of: https://github.com/brave/browser-ios/pull/1742

PR #1742, opened 2018-09-27
Setup script improvements

jhreis:
This also fixes the issue where the setup script would bail early and not correctly setup the host information for URP.


jhreis:
Thanks for fixing this @iccub!

jhreis:
@iccub Nope, the script already does this (you write it 😝)

I started https://github.com/brave/browser-ios/pull/1734 because my script was broken due to fabric/mixpanel removals, so it was never updating the host.

PR #1750, opened 2018-10-04
Bring back color attribute to Domain model.

iccub:


iccub:
One note, model files were sitting at root directory instead of `brave/src/data`, I moved them in this commit https://github.com/brave/browser-ios/pull/1750/commits/0f221bbae59f475e1e19ae11d365d8d4fa673840

I didn't copy them properly while porting the new cd changes I think

PR #1756, opened 2018-10-08
Save resolved bookmarks from sync.

iccub:
Without setting it to true, bookmarks are not saved and showed as added. Sync on server side seems to work fine.

### DIFF
--- a/brave/src/sync/Sync.swift
+++ b/brave/src/sync/Sync.swift
@@ -448,7 +448,7 @@ extension Sync {
                     
                 // TODO: Needs favicon
                 if clientRecord == nil {
-                    _ = recordType.coredataModelType?.add(rootObject: fetchedRoot, save: false, sendToSync: false, context: context)
+                    _ = recordType.coredataModelType?.add(rootObject: fetchedRoot, save: true, sendToSync: false, context: context)
                 } else {
                     // TODO: use Switch with `fallthrough`
                     action = .update


brave/src/sync/Sync.swift
```diff
@@ -448,7 +448,7 @@ extension Sync {
                     
                 // TODO: Needs favicon
                 if clientRecord == nil {
-                    _ = recordType.coredataModelType?.add(rootObject: fetchedRoot, save: false, sendToSync: false, context: context)
+                    _ = recordType.coredataModelType?.add(rootObject: fetchedRoot, save: true, sendToSync: false, context: context)
```
jhreis: I think this should be `false`, we should save after the parent `for-loop` is complete. Otherwise this will execute thousands of save operations super quickly.

iccub: You are correct, It didn't save because we missed passing background context in one place, will push a fix in a moment.

brave/src/data/Bookmark.swift
```diff
@@ -147,7 +147,8 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
     }
 
     static func add(rootObject root: SyncRecord?, save: Bool, sendToSync: Bool, context: NSManagedObjectContext) -> Syncable? {
-        add(rootObject: root as? SyncBookmark, save: save, sendToSync: sendToSync, parentFolder: nil)
+        add(rootObject: root as? SyncBookmark, save: save, sendToSync: sendToSync,
```
jhreis: If indenting some, should indent all parameters.

brave/src/data/Bookmark.swift
```diff
@@ -157,8 +158,8 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
     
     // Should not be used for updating, modify to increase protection
     class func add(rootObject root: SyncBookmark?, save: Bool = false, sendToSync: Bool = false,
-                   parentFolder: Bookmark? = nil) -> Bookmark? {
-        let context = DataController.newBackgroundContext()
+                   parentFolder: Bookmark? = nil,
+                   context: NSManagedObjectContext = DataController.newBackgroundContext()) -> Bookmark? {
```
jhreis: Same, all or non-indentation.

PR #1758, opened 2018-10-11
Update sync library to newest version.

iccub:


iccub:
This also requires to rerun `.setup.sh` script.

PR #1761, opened 2018-10-12
Use syncOrder algo for ordering favorites + minor sync-reorder fixes

iccub:


brave/src/data/Bookmark.swift
```diff
@@ -274,17 +274,28 @@ public final class Bookmark: NSManagedObject, WebsitePresentable, Syncable, CRUD
...
-        }
+        if dest === src { return }
         
         // Favorites use regular ordering algorithm.
         if src.isFavorite {
             reorderFavorites(frc: frc, sourceBookmark: src, destinationBookmark: dest,
                              sourceIndexPath: sourceIndexPath, destinationIndexPath: destinationIndexPath)
-            return
+        } else {
+            // Regular bookmarks use syncOrder based algorithm even if they are not in sync group.
```
jhreis: Maybe we should modify Favorites to use the new sync order too? May be best to keep them uniform where possible. How difficult would this be?

iccub:
Favorites ordering using syncOrder should work now. It required only few tweaks.

PR #1778, opened 2018-10-25
Move URP url from setup script to initialization in code.

iccub:


brave/src/analytics/user-referral-program/UserReferralProgram.swift
```diff
@@ -6,13 +6,13 @@ import Shared
...
 
 class UserReferralProgram {
-    private static let hostPlistKey = "URP_HOST"
     private static let apiKeyPlistKey = "API_KEY"
 
     private static let urpDateCheckPrefsKey = "urpDateCheckPrefsKey"
     private static let urpRetryCountdownPrefsKey = "urpRetryCountdownPrefsKey"
 
-    private static let stagingUrl = "https://laptop-updates-staging.herokuapp.com"
+    private let stagingUrl = "https://laptop-updates-staging.herokuapp.com"
```
jhreis: Maybe a struct:

```
private struct URL
    staging
    production
```

This will be ported, so may as well make it nice 😄

PR #1791, opened 2018-11-08
Update DAU weekly ping logic.

iccub:


brave/src/analytics/DAU.swift
```diff
@@ -166,12 +178,15 @@ struct DAU {
...
         let _year = stat[2]
         let SECONDS_IN_A_DAY = 86400
-        let SECONDS_IN_A_WEEK = 7 * 86400
         
         // On first launch, the user is all three of these
         let daily = dSecs >= SECONDS_IN_A_DAY
-        let weekly = dSecs >= SECONDS_IN_A_WEEK
+        
+        let weeksMonday = prefs.stringForKey(DAU.lastPingFirstMondayKey)
+        let weekly = (todayComponents.weeksMonday != weeksMonday) || weeksMonday == nil
```
jhreis: I'd get rid of parens and also maybe add comment as to why we are accounting for `weeksMonday` being `nil`.

PR #1805, opened 2018-11-14
Fix #1803, adds migration between old and new core data stacks.

### LINKED ISSUE #1803: update to 1.6.6 wipes out data
### Description:
Updating from 1.6.5 (app store version) to 1.6.6 via provided IPA file wipes out:
- Favorites
- Bookmarks
- History
- Session (opened tabs)

The follow appear to be retained:
- Stats
- Settings (specifically checked default search engine)
- Passwords

### Steps to Reproduce

  1. Download 1.6.5 from app store
  2. Add bookmarks, history, tabs, stats, cookies, logins/passwords, site specific shield settings, custom sites to ‘favorites/top site tiles’, set default search engine to DDG
  3. Use IPA file to update to 1.6.6

**Actual result:**
Be sad when your hard work is wiped out
image from ios

**Expected result:**
All data should be retained

easily

**Brave Version:**
1.6.5 -> 1.6.6 update

**Device details:**
reproduced on devices running iOS 10, 11, 12

**Website problems only:**
- did you check with Brave Shields down? n/a

- did you check in Puffin/Dolphin (UIWebView-based browsers)? n/a

- did you check in Safari/Firefox (WkWebView-based browsers)? n/a

### Additional Information
Reproduced by @kjozwiak @srirambv 

@srirambv also mentioned seeing a 'cleaning' message when updating on iOS12. I did not see this on iOS11. 

screen shot 2018-11-12 at 2 49 19 pm

iccub:
The problem was with removing persistent stores in new CD stack. It used a default store which in practice is a different database file. 

I brought back the persistent store that was used before new CD changes. In fact, there were two stores added `Brave.sqlite` and `Model.sqlite`, but from my tests I noticed that we only use the first one(despite comments saying that it was the old database file)

Now the app update should work correctly, tested it on development builds.

For more info, take a look at the old implementation of adding persistent store around this line https://github.com/brave/browser-ios/blob/6d9f4c7ddfacf74b5fbe87a2b68f50740e01db6a/brave/src/data/DataController.swift#L81

### DIFF
--- a/brave/src/data/DataController.swift
+++ b/brave/src/data/DataController.swift
@@ -9,6 +9,7 @@ private let log = Logger.browserLogger
 
 public class DataController: NSObject {
     public static var shared: DataController = DataController()
+    private let databaseName = "Brave.sqlite"
     
     private lazy var container: NSPersistentContainer = {
         let modelName = "Model"
@@ -21,12 +22,7 @@ public class DataController: NSObject {
         
         let container = NSPersistentContainer(name: modelName, managedObjectModel: mom)
         
-        if AppConstants.IsRunningTest {
-            let description = NSPersistentStoreDescription()
-            description.type = NSInMemoryStoreType
-            
-            container.persistentStoreDescriptions = [description]
-        }
+        addPersistentStore(for: container)
         
         // Dev note: This completion handler might be misleading: the persistent store is loaded synchronously by default.
         container.loadPersistentStores(completionHandler: { _, error in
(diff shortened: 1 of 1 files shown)


iccub:
This also needs to be ported to 1.7

jhreis:
We used to have a `Model.sqlite` but removed it, so it is there to handle migrations for those old users. That being said, we do not need to account for this anymore.

iccub:
Closing this since we are not going to use the new core data stack on 1.6 codebase.

